PR #16, opened 2022-05-05
feat: Add integration test for mongo db

WLefever-Cegeka:
Added integration test for MongoDB

### DIFF
--- a/ldes-server-infra-mongo/mongo-repository-library/pom.xml
+++ b/ldes-server-infra-mongo/mongo-repository-library/pom.xml
@@ -37,6 +37,15 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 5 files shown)


Yalz:
Just to be clear, does this test spin up a local mongodb instance, or this to be provided before running the tests?

WLefever-Cegeka:
> Just to be clear, does this test spin up a local mongodb instance, or this to be provided before running the tests?

This will spin up an in memory mongodb instance for integration tests. In this way you can verify that `LdesFragmentMongoRepository` effectively saves something in a mongodb.

.github/workflows/push-build-and-test.yml
```diff
@@ -2,7 +2,8 @@ name: Build & Test
 
 on:
   pull_request:
-    types: [ opened, edited, reopened ]
+    branches:
+      - master
```
Yalz, line 6: we'll probably need to change this to main, but tbd later

Yalz (approved):

PR #27, opened 2022-05-18
feat: Updated CODEOWNERS

WLefever-Cegeka:
Added Krieke to CODEOWNERS


Yalz (changes requested):
Would only add developers to code owners

PR #28, opened 2022-05-19
feat: Added FragmentProvider and LdesFragmentController

WLefever-Cegeka:
Added FragmentProvider and implementation.
Added LdesFragmentController.

Will add tests, documentation and correct containerization in follow-up PR.


ldes-server-port-publication-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/converters/JsonLDConverter.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.http.converter.HttpMessageNotWritableException;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.List;
+
+public class JsonLDConverter implements HttpMessageConverter<LdesFragment> {
```
sverholen, line 18: This is an example of why I prefer the old java-style convention for class naming over the Spring one. This class should be named JsonLdConverter according to the convention we're working with, as opposed to the old style JSONLDConverter, but not a mix of the two. Please rename.

WLefever-Cegeka: Updated name to JsonLdConverter

ldes-server-application/src/main/resources/application.yml
```diff
@@ -2,4 +2,8 @@ spring:
   data:
     mongodb:
       uri: mongodb+srv://${MONGO_USERNAME}:${MONGO_PASSWORD}@${MONGO_URL}/${MONGO_DATABASE}
-      database: ${MONGO_DATABASE}
\ No newline at end of file
+      database: ${MONGO_DATABASE}
+
+ldes:
```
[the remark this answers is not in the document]
WLefever-Cegeka: Totally agree, would it be okay if I do this in a follow up PR? I mentioned that in the initial message: "Will add tests, documentation and correct containerization in follow-up PR."

sverholen (approved):

PR #45, opened 2022-05-30
feat: docker-compose for server (and minor improvements)

sverholen:


docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN git clone ${DOCKER_LDES_BOM_REPO} bom
+
+# MAVEN: application
+FROM builder as app-stage
+
+COPY --from=bom-stage bom/pom.xml .
+RUN mvn install
+WORKDIR /
+COPY . /
+RUN mvn install
```
WLefever-Cegeka, line 21: This is incorrect. Since you are in the folder `docker`, you copy the contents of the folder `docker`, which is the `Dockerfile`. To solve this issue you could either clone the REPO or move de Dockerfile one level higher (this is needed as the Dockerfile can only COPY on its own level and below).
There's also no distinction between line 18 and line 21. These install the bom twice.

docker-compose.yml
```diff
@@ -2,23 +2,25 @@ version: '3.3'
 services:
   ldes-server:
     env_file:
-      - ./config.local.env
-    container_name: ldes-server
+      - ./docker-compose/config.local.env
+    container_name: ldes.server
```
WLefever-Cegeka, line 6: out of curiosity, what's the reasoning behind the dot?

WLefever-Cegeka (changes requested):

WLefever-Cegeka:
Thanks for the work regarding containerization and the documentation!

docker-compose.yml
sverholen: The dot is irrelevant, but the containers should have names that distinguishes them from containers that use the same image in other projects (outside of vsds). Do you prefer an underscore ?

docker/Dockerfile
sverholen: The context is derived from the context declaration in docker-compose.yml.
The first mvn install installs the BOM in the local repo, the COPY makes sures that the second mvn install installs the app

docker-compose.yml
WLefever-Cegeka: no, it's okay :-)

docker/Dockerfile
WLefever-Cegeka: Aha, indeed. Tried building the container in the `docker` folder. In that sense, it would be safer to clone the repo, but can agree with this as well.

WLefever-Cegeka (approved):

PR #82, opened 2022-06-21
feat: Update method name to 'retrieveLdesFragments'

WLefever-Cegeka:


Yalz (commented):
Since we are retrieving 1 Ldes fragment, we should best rename the method to retrieveLdesFragment

PR #86, opened 2022-06-27
fix: SonarCloud reports

WLefever-Cegeka:


ldes-server-port-publication-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/constants/PublicationRestTestConstants.java
```diff
@@ -0,0 +1,12 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.rest.constants;
+
+public final class PublicationRestTestConstants {
+
+    public static final String FRAGMENTATION_VALUE_1 = "2020-12-28T09:36:09.72Z";
+
+     private PublicationRestTestConstants() {
+        // No need to instantiate the class, since there are only constants in it
```
Yalz, line 8: Although the comment clarifies it, I'd delete it since it clutters the code.

Yalz (commented):
In general, I would rediscuss the approach on test constants. Whilst I can understand why you are not a fan of a shared testconstants class over the modules, it now looks a bit too scattered too me.

ldes-server-port-publication-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/constants/PublicationRestTestConstants.java
WLefever-Cegeka: I think that will address all your comments :-D

Yalz (approved):

PR #87, opened 2022-06-27
fix: remove the LdesMember contructor, only used in tests

WLefever-Cegeka:
This constructor is only used in tests. Would rather see it removed and do the necessary conversion from String to Model in our tests.

```
    public LdesMember(final String ldesMember, Lang lang) {
        memberModel = RDFParserBuilder.create().fromString(ldesMember).lang(lang).toModel();
    }
```


Yalz (commented):
Although I agree with you on removing the constructor since we only use this in test, duplicating this method in every test class we use it seems a bit off to me as well..

Yalz (approved):

PR #90, opened 2022-06-28
Fix/revise add member method

WLefever-Cegeka:
This is mainly an effort in restructuring the code. I haven't fixed existing issues to keep the PR size limited.
Note that at this moment there are still the following issues:

- The new fragment we create is not based on the LDESMember
- When there are more members with the same value for path, we will create a new fragment with the same Id (error)
- When using the parent url (http://localhost:8080/mobility-hindrances) the first fragment is not shown.

### DIFF
--- a/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/FragmentInfo.java
+++ b/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/FragmentInfo.java
@@ -3,26 +3,18 @@ package be.vlaanderen.informatievlaanderen.ldes.server.domain.entities;
 public class FragmentInfo {
     private final String view;
     private final String shape;
-    private final String parentFragmentId;
     private final String viewShortName;
     private final String path;
     private String value;
-    private Long memberLimit;
     private Boolean immutable;
 
-    public FragmentInfo(String view, String shape, String parentFragmentId, String viewShortName, String path, String value, Long memberLimit) {
+    public FragmentInfo(String view, String shape, String viewShortName, String path, String value) {
         this.view = view;
         this.shape = shape;
-        this.parentFragmentId = parentFragmentId;
         this.viewShortName = viewShortName;
         this.path = path;
         this.value = value;
-        this.memberLimit = memberLimit;
         this.immutable = false;
-    }
-
-    public String getParentFragmentId() {
-        return parentFragmentId;
     }
 
     public String getPath() {
@@ -35,14 +27,6 @@ public class FragmentInfo {
 
     public void setValue(String value) {
         this.value = value;
-    }
-
-    public Long getMemberLimit() {
-        return memberLimit;
-    }
-
-    public void setMemberLimit(Long memberLimit) {
-        this.memberLimit = memberLimit;
     }
 
     public String getViewShortName() {
--- a/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/TreeRelation.java
+++ b/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/TreeRelation.java
@@ -1,8 +1,6 @@
 package be.vlaanderen.informatievlaanderen.ldes.server.domain.entities;
 
-import org.apache.jena.rdf.model.Literal;
-import org.apache.jena.rdf.model.Resource;
-import org.apache.jena.rdf.model.ResourceFactory;
+import java.util.Objects;
 
 public class TreeRelation {
     private String treePath;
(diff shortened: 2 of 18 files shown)


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/TreeRelation.java
```diff
@@ -1,38 +1,46 @@
...
 
 public class TreeRelation {
-    private String treePath;
-    private String treeValue;
-    private String treeNode;
-    private String rdfSyntaxType;
+    private final String treePath;
+    private final String treeValue;
+    private final String treeNode;
+    private final String rdfSyntaxType;
```
Yalz, line 9: Did you test this? having this final and not having a default contructor gave an issue with MongoDB

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/repositories/LdesFragmentRespository.java
```diff
@@ -1,14 +1,13 @@
...
-import be.vlaanderen.informatievlaanderen.ldes.server.domain.config.ViewConfig;
 import be.vlaanderen.informatievlaanderen.ldes.server.domain.entities.LdesFragment;
 
 import java.util.Optional;
 
 public interface LdesFragmentRespository {
     LdesFragment saveFragment(LdesFragment ldesFragment);
 
-    Optional<LdesFragment> retrieveFragment(String viewShortName, String path, String value);
+    Optional<LdesFragment> retrieveFragmentByViewShortNameAndPathAndType(String viewShortName, String path, String value);
```
Yalz, line 10: Not sure if this improves the readability

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/services/FragmentationServiceImpl.java
```diff
@@ -24,58 +19,42 @@ public class FragmentationServiceImpl implements FragmentationService {
...
-                    }
-                    else {
-                        return fragment;
-                    }
-                })
-                .orElse(makeNewFragment(ldesMember));
-        addLdesMembertoLdesFragment(ldesMember, ldesFragment);
+        ldesMember.replaceTreeMemberStatement(ldesConfig.getHostName(), ldesConfig.getCollectionName());
+        LdesFragment ldesFragment = retrieveLastFragmentOrCreateNewFragment(ldesMember);
+        ldesFragment.addMember(ldesMember);
```
Yalz, line 37: This seems to not take into account if you have to shift members in a fragment (to a new fragment)

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/services/TimeBasedFragmentCreator.java
```diff
@@ -0,0 +1,49 @@
...
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.entities.LdesFragment;
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.entities.LdesMember;
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.entities.TreeRelation;
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.repositories.LdesFragmentRespository;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+
+@Component
+public class TimeBasedFragmentCreator implements FragmentCreator {
```
Yalz, line 15: I do like this approach immediately ! In the future we might need to extract this to a separate module, but great we already start making an interface!

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/services/TimeBasedFragmentCreator.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private void makeFragmentImmutableAndUpdateRelations(LdesFragment completeLdesFragment, LdesFragment newFragment) {
+        completeLdesFragment.setImmutable(true);
+        completeLdesFragment.addRelation(new TreeRelation(newFragment.getFragmentInfo().getPath(), newFragment.getFragmentId(), newFragment.getFragmentInfo().getValue(), TREE_GREATER_THAN_RELATION));
+        ldesFragmentRespository.saveFragment(completeLdesFragment);
+        newFragment.addRelation(new TreeRelation(completeLdesFragment.getFragmentInfo().getPath(), completeLdesFragment.getFragmentId(), completeLdesFragment.getFragmentInfo().getValue(), TREE_LESSER_THAN_RELATION));
+    }
+
+    private LdesFragment createFreshFragment(LdesMember firstMember) {
```
Yalz, line 44: Would propose a different name for this. Would keep createNewFragment and maybe the other one with a different name?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/FragmentInfo.java
```diff
@@ -3,28 +3,20 @@
 public class FragmentInfo {
     private final String view;
     private final String shape;
-    private final String parentFragmentId;
     private final String viewShortName;
     private final String path;
     private String value;
-    private Long memberLimit;
```
Yalz, line 10: What is the reasoning for removing the memberlimit and the parent? 
The parent is part of the design for the future, but I can understand why you prefer to only introduce that whilst we need it.
I presume you removed the limit since you only use the viewconfig?

Yalz (approved):
I think the split up towards Time based fragmentation is a good one! 
My only concern now is how we plan to incorporate more fragmentations in this design ? 
(Not specific to this PR, but food for thought)

WLefever-Cegeka:
> I think the split up towards Time based fragmentation is a good one! My only concern now is how we plan to incorporate more fragmentations in this design ? (Not specific to this PR, but food for thought)

Ha, I did think about this.
In spring you can use the Qualifier annotation to indicate based on configuration which implementation of an interface you want to use. Using this annotation together with loading beans would do the trick :-)
For the moment I added the default annotation on TimeBasedFragmentCreator

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/TreeRelation.java
WLefever-Cegeka: Reverted this. Didn't realize that this domain class was also used as part of the MongoDB entity.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/repositories/LdesFragmentRespository.java
WLefever-Cegeka: Ha, totally debatable. Usually, I find method names like `retrieveFragment` a bit to vague and like to add some more detail e.g. the argument which is used for selecting the fragment, but I also understand your point of view, ``retrieveFragmentByViewShortNameAndPathAndType` is a pretty long name :-)

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/services/TimeBasedFragmentCreator.java
WLefever-Cegeka: Top!

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/entities/FragmentInfo.java
WLefever-Cegeka: Yep, those were my concerns :-)

PR #93, opened 2022-06-29
docs: updating README

Yalz:


README.md
```diff
@@ -8,35 +8,23 @@ The LDES Server is a configurable component that can be used to ingest, store, t
 We'll show you how to set up your own LDES Server both locally and via Docker using a mongo db for storage and a rest endpoint for ingestion and publication.
 Afterwards, you can change storage, ingestion and publication options by plugging in other components.
 
-### Locally
+## Locally
 
-#### Docker-compose
+### Docker-compose
 
 There are 3 files where you can configure the dockerized application:
```
Krieke, line 15: I see only 2 mentioned. The config file and the docker compose files.

README.md
```diff
@@ -8,35 +8,23 @@ The LDES Server is a configurable component that can be used to ingest, store, t
...
-# The LDES BOM repository
-DOCKER_LDES_BOM_REPO=https://github.com/Informatievlaanderen/VSDS-LDES.git
-# Run the LDES server app on this port
-DOCKER_LDES_SERVER_PORT=6060
-```
-
-##### <a name="docker-compose-config-config"></a>The config files: [config.env](docker-compose/config.env) or [config.local.env](docker-compose/config.local.env)
+#### <a name="docker-compose-config-config"></a>The config files: [config.env](docker-compose/config.env) or [config.local.env](docker-compose/config.local.env)
 
 Runtime settings can be defined in the configuration files. Use [config.env](docker-compose/config.env) for a public setup and be sure not to commit this file, as it contains secrets. For a local setup, use [config.local.env](docker-compose/config.local.env). It's safe to commit this file with secrets, since they'll only be relevant to your local environment (assuming you don't reuse personal passwords for public services for your local services).
```
Krieke, line 21: It is never safe to commit files with secrets. But probably I am mis understanding the concept of local commit. Local creation without this file being part of git I might understand.

Krieke: Should it then be a good practice to add the config.local.env and config.env to the .gitignore file?

README.md
```diff
@@ -45,15 +33,47 @@ COMPOSE_DOCKER_CLI_BUILD=1 docker-compose build
 docker-compose up
 ```
 
-The server application will be available on the port you configured with `DOCKER_LDES_SERVER_PORT` (or port 8080 by default).
-
-#### IntelliJ
+### Maven
```
Krieke, line 36: It would be good to extend this with:
* build locally
* run unit tests with and without coverage
* run integration tests with and without coverage
* run all tests with and without coverage
* and how to run it => but that was already mentioned

README.md
```diff
@@ -45,15 +33,47 @@ COMPOSE_DOCKER_CLI_BUILD=1 docker-compose build
...
 
-Now you only need to provide the correct environment variables by overriding the placeholders in [application.yml](ldes-server-application/src/main/resources/application.yml).
+```mvn
+mvn spring-boot:run
+```
 
-And you can spin up your own LDES Server.
\ No newline at end of file
+This will run a clean slate LDES Server depending on your needs, one or of the following maven profiles can be activated:
+
```
Krieke, line 45: Can you mention were people can find this application config => so the pom.xml. Think like not everyone has java / maven knowledge :-)

Krieke (commented):
These are some of the items I would add. It mostly involves being a bit more informative on the java / maven steps for those none java people who also want to use are extend this repository.

README.md
Krieke: Maybe explain a bit more about the configuration options in these config files. Some are self explainable others or not. 
So what happens if no LDES configuration is set in the config.local.env?  Apparently also defaults are used but it is not clear which they are from within the readme, config files

README.md
Yalz: Indeed. The docker part needs to still be looked at.

README.md
Yalz: This could be added, but can we not assume that anyone who stumbles upon a maven/java/spring project will either have this knowledge or first get familiar with the technology?

Krieke: I see what you mean... maybe create a ref link `application config` which is mapped to the pom.xml  => is min effort :-)

README.md
Yalz: To me, it boils down back to the familiarity of the user with maven. 
Since all operations you mentioned have their maven lifecycle (mvn install, mvn verify, mvn test)
Considering coverage, should we provide local coverage? We'll include the sonar badges that indicate our coverage , code smells, ... but I don't see why a developer wants to manually check coverage themselves? (at least not in a scenario where we need to assist him in this)

Krieke: I understand, but it doesn't take that much effort to add it so it is clear for QA people. The above is actually a min set of documentation that I add mostly in my README.md, how to build, how to test, how to dockerize.

With respect to the code review this is currently not supported by the SonarCloud that is used for this repository. Also the code coverage it not only about a number but actually checking which code parts are still not hit. 80% is a threshold defined number, but the test should mostly cover certain pieces of code. This you do check as a developer with manual checks imho. 

Maybe we have to discus with other stakeholders on this matter about the few mvn instructions to document for not experience java / mvn users.  Because in the end you might also say... read the code to understand this component :-)

sverholen (approved):

PR #95, opened 2022-06-29
fix: remaining code smells

WLefever-Cegeka:


ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/LdesFragmentMongoRepository.java
```diff
@@ -31,6 +32,6 @@ public Optional<LdesFragment> retrieveLastFragment(String view) {
         return repository.findAll().stream()
                 .filter(ldesFragmentEntity -> !ldesFragmentEntity.isImmutable())
                 .map(LdesFragmentEntity::toLdesFragment)
-                .findFirst();
+                .min(Comparator.comparing(LdesFragment::getFragmentId));
```
Yalz, line 35: Seems like my comments would also fit the previous version too, but isn't this the opposite of what our method says?

I'd expect that we sort on the fragment value in descending order?

ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/entities/LdesFragmentEntity.java
```diff
@@ -10,7 +10,7 @@
 import java.util.List;
 
 @Document("ldesfragment")
-public class LdesFragmentEntity implements Comparable<LdesFragmentEntity> {
```
Yalz, line 13: not sure why you'd want to remove this?

ldes-server-port-ingestion-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingestion/rest/LdesMemberIngestionController.java
```diff
@@ -16,7 +16,6 @@ public LdesMemberIngestionController(final FragmentationService ldesReader) {
         this.ldesReader = ldesReader;
     }
 
-    // TODO update value to an injected value
     @PostMapping(value = "/ldes-member", consumes = "application/n-quads")
```
Yalz, line 19: The main point for this TODO was to maybe replace this with /{shortname} . Is this something that we might want?

ldes-server-port-publication-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/LdesFragmentControllerTest.java
```diff
@@ -71,33 +68,36 @@ void when_GETRequestIsPerformed_ResponseContainsAnLDesFragment(String mediaType,
...
                 .toModel();
-        //TODO
-//        assertTrue(resultModel.contains(createStatement(createResource(ldesFragmentConfig.getView()),
-//                createProperty("https://w3id.org/tree#shape"), createResource(ldesFragmentConfig.getShape()))));
+        assertEquals("https://private-api.gipod.test-vlaanderen.be/api/v1/ldes/mobility-hindrances/shape", getObjectURI(resultModel, RdfContants.TREE_SHAPE));
+        assertEquals("http://www.w3.org/ns/prov#generatedAtTime", getObjectURI(resultModel, RdfContants.LDES_TIMESTAMP_PATH));
+        assertEquals("http://purl.org/dc/terms/isVersionOf", getObjectURI(resultModel, RdfContants.LDES_VERSION_OF));
+        assertEquals("http://localhost:8080/mobility-hindrances?generatedAtTime=2020-12-28T09:36:09.72Z", getObjectURI(resultModel, RdfContants.TREE_VIEW));
+        assertEquals("https://w3id.org/ldes#EventStream", getObjectURI(resultModel, RdfContants.RDF_SYNTAX_TYPE));
         verify(fragmentationService, times(1)).getFragment(ldesConfig.getCollectionName(), viewConfig.getTimestampPath(), FRAGMENTATION_VALUE_1);
```
Yalz, line 76: I would if possible extract these "Magic strings" to the top?

ldes-server-port-publication-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/LdesFragmentControllerTest.java
```diff
@@ -108,4 +108,17 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
...
+    static class TestConfig {
+        @Bean
+        public ViewConfig viewConfig(){
+            ViewConfig viewConfig = new ViewConfig();
+            viewConfig.setShape("https://private-api.gipod.test-vlaanderen.be/api/v1/ldes/mobility-hindrances/shape");
+            viewConfig.setMemberLimit(3L);
+            viewConfig.setTimestampPath("http://www.w3.org/ns/prov#generatedAtTime");
+            viewConfig.setVersionOfPath("http://purl.org/dc/terms/isVersionOf");
+            return viewConfig;
+        }
```
Yalz, line 121: Is there no way to use the application-test.yml in this instance? or do you specifically prefer to have the testconfig inside this test class?

Yalz (approved):

ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/LdesFragmentMongoRepository.java
WLefever-Cegeka: Indeed, we should go with something like retrieveFirstOpenFragment()
I'll add a test :-) That should clarify the behaviour.

ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/entities/LdesFragmentEntity.java
WLefever-Cegeka: this requires the method compareTo to be overriden.
And according to java:S1210: "It is strongly recommended, but not strictly required that (x.compareTo(y)==0) == (x.equals(y))."

In the current implementation the compareTo method only uses the fragmentId
In our case we don't really that equality only depends on the fragmentId. (because that would mean that a fragment fragments can have different relations, members, ... and still be equal)

Hence, I removed the compareTo method, the Comparable interface and replaced the code usage of the compareTo with 
```
.min(Comparator.comparing(LdesFragment::getFragmentId))
```

Yalz: Well in theory we should have compared on the value of the path for a fragment. but the current approach is also fine for me now

ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/LdesFragmentMongoRepository.java
Yalz: Well i think we want to retrieve the last fragment no? Need to add more test cases indeed

WLefever-Cegeka: The more I think about it first and last don't really mean anything. In time based fragmentation those words have a value, but in a general LDES one would expect that based on the fragmentation there's only one open fragment. Called the method retrieveOpenFragment() for now. And added tests ;-)

PR #97, opened 2022-06-30
Fix/vsdspub 143 fix docker

WLefever-Cegeka:


.github/workflows/pr-merged.yml
```diff
@@ -54,4 +54,4 @@ jobs:
         with:
           context: .
           push: true
-          tags: ${{ env.version }}-${{steps.current-time.outputs.formattedTime }}
\ No newline at end of file
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.version }}-${{steps.current-time.outputs.formattedTime }}
```
Yalz, line 57: As a docker novice, how would this tag translate?

PR #102, opened 2022-07-01
fix: VSDSPUB-62: cache control

WLefever-Cegeka:


Krieke:
@rorlic There was one code smell reported. By not fixing it now, it might be not visible the next time, but still be an issue that is seen only when logged into SonarCloud directly. This is purely a fyi

PR #118, opened 2022-07-13
Feature/fragmentation

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/converter/FragmentIdConverter.java
```diff
@@ -1,15 +1,20 @@
...
-    }
-
-    private static String getCompactTimestampPath(String timestampPath) {
-        return timestampPath.split("#")[1];
+    public static String toFragmentId(String hostLocation, String view, List<FragmentPair> fragmentPairs) {
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("%s/%s".formatted(hostLocation, view));
+        if(!fragmentPairs.isEmpty()){
+            stringBuilder.append("?");
+            fragmentPairs.forEach(fragmentPair -> stringBuilder.append("%s=%s".formatted(fragmentPair.fragmentKey(), fragmentPair.fragmentValue())));
```
Yalz, line 16: Just as a general point, do we plan to limit what can be the fragmentvalue? I am picturing a string value that contains '"' or '=' which could break the url

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/FragmentationServiceImpl.java
```diff
@@ -24,18 +28,18 @@ public FragmentationServiceImpl(LdesConfig ldesConfig, ViewConfig viewConfig,
     @Override
     public LdesFragment getFragment(String collectionName, String path, String value) {
         return ldesFragmentRespository.retrieveFragment(collectionName, path, value)
-                .orElseGet(()-> createDummyFragment(collectionName, path));
+                .orElseGet(()-> createDummyFragment(collectionName, List.of(new FragmentPair(path, value))));
```
Yalz, line 31: I'm wondering if 'Dummy' is the right word compared to 'EmptyFragment' for example. But I presume it boils down to taste :)

Yalz (approved):
Great work! will clean up a hard coded way of thinking of the timestamppath!

Yalz:
Don't forget to check the readme for updates. I think we mention the endpoint for publication in there. now with the new fragmentPair, this might need an update?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/FragmentationServiceImpl.java
WLefever-Cegeka: renamed dummy to empty

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/converter/FragmentIdConverter.java
WLefever-Cegeka: Hmm, that's correct. Will add an issue so that we won't forget to tackle this in a follow up PR.

WLefever-Cegeka:
> Don't forget to check the readme for updates. I think we mention the endpoint for publication in there. now with the new fragmentPair, this might need an update?

Updated README.md, so that we only mention the base url: /{ldes.collection-name}

PR #121, opened 2022-07-14
Feature/vsdspub 216 performance improvements

WLefever-Cegeka:


ldes-server-infra-mongo/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/LdesMemberMongoRepository.java
```diff
@@ -36,4 +38,9 @@ public Optional<LdesMember> getLdesMemberById(String memberId) {
                 .map(LdesMemberEntity::toLdesMember);
     }
 
+    @Override
+    public Stream<LdesMember> getLdesMembersByIds(List<String> ids) {
+        return StreamSupport.stream(repository.findAllById(ids).spliterator(), false).map(LdesMemberEntity::toLdesMember);
```
Yalz, line 43: how are unfindable ids handled?

Yalz (approved):
I can see why this is better performance wise, but i'm just not sure if we're not neglecting the scenarios where a member doesn't exist for some reason.

sverholen (approved):

PR #124, opened 2022-07-18
feat: make member id predicate configurable

Yalz:


sverholen (approved):
IMO:
- don't pass membertype everywhere
- separate fragment and member logic in memberingestservice

sverholen (approved):

PR #206, opened 2022-09-07
Feature/split repository member fragment

WLefever-Cegeka:


Yalz (changes requested):
I do not see any point why we would need this now, as this, IMO, adds an extra complexity to our code without benefit.

PR #256, opened 2022-09-30
Feature: Move child creation into domain object

WLefever-Cegeka:


Yalz (approved):

Yalz:
Maybe just write an extra test for GeospatialFragmentationStrategy ?

WLefever-Cegeka:
> Maybe just write an extra test for GeospatialFragmentationStrategy ?

Good catch, working on that in #258

PR #319, opened 2022-12-05
Added Pieter-Jan Livens to CODEOWNERS

pj-cegeka:


Yalz:
Would maybe good to also remove tom? since he's not an active reviewer?

sverholen (approved):

PR #369, opened 2023-01-04
Feat/auto add issues and forks to project

Yalz:


WLefever-Cegeka (changes requested):
Is there a need for both this and https://github.com/Informatievlaanderen/VSDS-LDESServer4J/blob/main/.github/workflows/add-fork-PR-to-project.yml

PR #377, opened 2023-01-06
Fix: Issue-376: empty accept header

WLefever-Cegeka:


ldes-server-port-fetch-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/eventstream/converters/EventStreamHttpConverter.java
```diff
@@ -39,7 +41,7 @@ public boolean canWrite(Class<?> clazz, MediaType mediaType) {
 
 	@Override
 	public List<MediaType> getSupportedMediaTypes() {
-		return List.of(MediaType.ALL);
+		return List.of(DEFAULT_MEDIA_TYPE, MediaType.ALL);
```
Yalz, line 44: Should this be needed? Since MediaType.ALL should contain "all" no ? :)

Yalz (approved):

ldes-server-port-fetch-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/eventstream/converters/EventStreamHttpConverter.java
WLefever-Cegeka: This is a Spring Boot issue. If no content-type is given, then MediaType.ALL uses application/json by default. In our case we want to use text/turtle as default. Hence the explicit DEFAULT_MEDIA_TYPE.

PR #392, opened 2023-01-16
Fix issue 281

Tomvbe:


WLefever-Cegeka (changes requested):
Hi Tom,
Overall code looks very good!
But would like to reflect a bit on certain aspects with you so indicate 'Request changes'.

WLefever-Cegeka (approved):

PR #396, opened 2023-01-18
fix/ parse string to duration from duration and period format

pj-cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/converter/DurationParser.java
```diff
@@ -0,0 +1,35 @@
...
+			return Duration.parse(duration);
+		} catch (DateTimeParseException ignored) {
+			// if text cannot be parsed to duration, try parsing it to period
+		}
+		try {
+			Period period = Period.parse(duration);
+			return periodToDuration(period);
+		} catch (DateTimeParseException e) {
+			throw new DurationParserException(duration);
+		}
```
WLefever-Cegeka, line 32: This could be placed inside the braces of the catch above.

WLefever-Cegeka (approved):

PR #422, opened 2023-02-09
Fix/update docs

Tomvbe:


README.md
```diff
@@ -85,21 +85,33 @@ cd ldes-server-application
...
 
 ```mvn
-mvn spring-boot:run -P {profiles (comma separated with no spaces) }
+mvn spring-boot:run -P{profiles (comma separated with no spaces) }
 ```
 
-This will start an empty LDES server. To enrich this server, certain Maven profiles can be activated:
+for example:
+```mvn
+mvn spring-boot:run -P{fragmentation-timebased,http-fetch,http-ingest,queue-none,storage-mongo}
```
Yalz, line 93: This approach will need to change in the future, but good for now :)

Yalz (approved):
You've only done it for the application pom, but you will need it in every pom no?

PR #433, opened 2023-02-16
Isolate TreeNodeInfoResponse

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/tree/node/services/TreeNodeConverterImpl.java
```diff
@@ -47,17 +48,14 @@ public Model toModel(final TreeNode treeNode) {
...
-					treeNode.getFragmentId()));
-		});
+		List<TreeRelationResponse> treeRelationResponses = treeNode.getRelations().stream()
+				.map(treeRelation -> new TreeRelationResponse(treeRelation.treePath(),
+						ldesConfig.getBaseUrl() + treeRelation.treeNode(),
+						treeRelation.treeValue(), treeRelation.treeValueType(), treeRelation.relation()))
+				.toList();
+		TreeNodeInfoResponse treeNodeInfoResponse = new TreeNodeInfoResponse(treeNode.getFragmentId(),
+				treeRelationResponses);
+		List<Statement> statements = new ArrayList<>(treeNodeInfoResponse.convertToStatements());
```
Tomvbe, line 58: Do we need to clone this list again if 'convertToStatements' already creates a new list?

Tomvbe (approved):

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/tree/node/services/TreeNodeConverterImpl.java
WLefever-Cegeka: Hi @Tomvbe,
Your remark made me think about something. One expects a set of statements from a treeNodeInfoResponse to be consistent and immutable, i.e. after a TreeNodeInfoResponse is converted to a set of statements it should be impossible to change or update the statements. So I added immutability, also to the other conversions.

PR #434, opened 2023-02-16
Use built-in prefix when possible

Tomvbe:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/converter/PrefixAdderImpl.java
```diff
@@ -9,24 +9,26 @@
...
+	private static final String VALID_PREFIX_REGEX = "^[a-zA-Z_][\\w.-]*$"; // NCName regex
 	private static final Logger LOGGER = LoggerFactory.getLogger(PrefixAdderImpl.class);
 
 	@Override
 	public Model addPrefixesToModel(Model model) {
 		Map<String, String> nameSpaceMap = new HashMap<>();
 		Map<String, String> localNamesMap = new HashMap<>();
 		model.listStatements().forEach(statement -> extractNamespaces(nameSpaceMap, localNamesMap, statement));
-		removePrefixesForWhichLocalNameDoesIsNotCompliant(nameSpaceMap, localNamesMap);
+		removePrefixesForWhichLocalNameIsNotCompliant(nameSpaceMap, localNamesMap);
```
Yalz, line 26: from a Clean Code perspective, I both love and hate this method name ..
It says exactly what it does, yet this name is too long in my opinion.. removeNonComplaintPrefixes perhaps?

Yalz (approved):

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/converter/PrefixAdderImpl.java
Tomvbe: The problem is that the prefix IS compliant, we just do not use a prefix if the namespace has one or more non compliant localNames.. I was able to cut out a couple of extra chars but could not come up with a name that fits better. Since it's a private method that is only called once, I would suggest to leave it as is for the time being.

PR #456, opened 2023-03-13
Feat/event stream testing

WLefever-Cegeka:


Yalz (approved):
Looking good! 
Please address the Sonarcloud bug first though

PR #480, opened 2023-03-29
Fix: Exclude mongock changeset from coverage

WLefever-Cegeka:


ldes-server-infra-mongo/mongock-changeset-1/pom.xml
```diff
@@ -34,13 +34,24 @@
...
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/mongock/changeset1/*</exclude>
```
Tomvbe, line 50: would it be possible to move the wildcard one level higher? Then all future changesets are excluded as well

Tomvbe (approved):

PR #481, opened 2023-03-30
Feat: Add versionOf and timestamp attribute to Member

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/viewcreation/valueobjects/ViewConfig.java
```diff
@@ -5,18 +5,42 @@
...
 
 @Configuration
 @ConfigurationProperties
 public class ViewConfig {
 
+	public static final String DEFAULT_VIEW_NAME = "by-page";
+	public static final String DEFAULT_VIEW_FRAGMENTATION_STRATEGY = "pagination";
+	public static final Map<String, String> DEFAULT_VIEW_FRAGMENTATION_PROPERTIES = Map.of("memberLimit", "100");
 	private List<ViewSpecification> views = new ArrayList<>();
+	private Boolean defaultView;
```
Tomvbe, line 18: can this be a primitive?

pj-cegeka (approved):

ldes-server-infra-mongo/mongo-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/member/MemberMongoRepository.java
```diff
@@ -30,14 +27,7 @@ public MemberMongoRepository(final LdesMemberEntityRepository repository, MongoT
...
 	public Member saveLdesMember(Member member) {
-		Query query = new Query();
-		query.addCriteria(Criteria.where("_id").is(member.getLdesMemberId()));
-		Update update = new Update();
-		StringWriter outputStream = new StringWriter();
-		RDFDataMgr.write(outputStream, member.getModel(), Lang.NQUADS);
-		String ldesMemberString = outputStream.toString();
-		update.set("model", ldesMemberString);
-		mongoTemplate.upsert(query, update, LdesMemberEntity.class);
+		repository.save(LdesMemberEntity.fromLdesMember(member));
```
Tomvbe, line 30: This is fine for me, but I thought we used the template for performance reasons?

WLefever-Cegeka: That's correct. Template was used so that the attributes of LdesMemberEntity could be updated separately (this enabled fragmentation to already start even if the member was not saved). The reason why I removed that here, is because we're going to rethink the whole ingest-fragmentation anyway and it feels icky to use mongoTemplate to update almost all attributes of a member at the same time (model, versionOf, timestamp, id).

PR #484, opened 2023-03-31
Feature/vsdspub 564 snapshot

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/snapshot/services/MemberReduceOperator.java
```diff
@@ -0,0 +1,16 @@
...
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.tree.member.entities.Member;
+
+import java.util.function.BinaryOperator;
+
+public class MemberReduceOperator implements BinaryOperator<Member> {
+	@Override
+	public Member apply(Member member, Member member2) {
+
+		if (member.getTimestamp().isAfter(member2.getTimestamp()))
```
Yalz, line 11: just to double check.. timestamp is a mandatory value on the member right?

Yalz (approved):
Looking good! Just to verify.. Since i found linking between Snapshot fragments, I presume we define a max size of a fragment?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/snapshot/services/MemberReduceOperator.java
WLefever-Cegeka: well, actually not. timestamppath is an optional attribute according to the ldes spec. I'll add validation of the members and exception handling in a follow-up PR on branch `feature/VSDSPUB-564-snapshot-exceptions`

PR #492, opened 2023-04-04
Feature/dockerfile update

WLefever-Cegeka:


Dockerfile
```diff
@@ -30,7 +28,5 @@ COPY --from=app-stage ldes-fragmentisers/ldes-fragmentisers-substring/target/lde
 COPY --from=app-stage ldes-fragmentisers/ldes-fragmentisers-pagination/target/ldes-fragmentisers-pagination-jar-with-dependencies.jar ./lib/
 COPY --from=app-stage ldes-queues/ldes-queue-none/target/ldes-queue-none-jar-with-dependencies.jar ./lib/
 
-RUN useradd -u 2000 ldes
```
Tomvbe, line 33: why are we removing this? won't this reduce security ?

Dockerfile
```diff
@@ -13,9 +13,7 @@ RUN mvn net.revelc.code.formatter:formatter-maven-plugin:format install -DskipTe
 #
 # RUN THE APPLICATION
 #
-FROM openjdk:18-ea-bullseye
-RUN apt-get update & apt-get upgrade
```
Tomvbe, line 17: why do we leave this out? This can also reduce security

Dockerfile
```diff
@@ -13,9 +13,7 @@ RUN mvn net.revelc.code.formatter:formatter-maven-plugin:format install -DskipTe
...
-
+FROM amazoncorretto:17-al2023-jdk
```
Tomvbe, line 16: We are now going from 18 to 17. Perhaps we could go up instead? Locally I run corretto 19.0.2 and this works fine with the server. That would be image 19.0.2-alpine3.17

Tomvbe:
replaced by https://github.com/Informatievlaanderen/VSDS-LDESServer4J/pull/969

PR #505, opened 2023-04-12
Feat/implement fetch flow

Tomvbe:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RootFragmentCreatorImpl.java
```diff
@@ -18,16 +17,13 @@ public RootFragmentCreatorImpl(LdesFragmentRepository ldesFragmentRepository) {
...
-		}
-		return optionalRoot.get();
+		return ldesFragmentRepository
+				.retrieveRootFragment(viewName)
+				.orElseGet(() -> createRoot(viewName));
 	}
 
 	private LdesFragment createRoot(String viewName) {
-		LdesFragment ldesFragment = new LdesFragment(viewName, List.of());
+		LdesFragment ldesFragment = new LdesFragment(viewName.split("/")[0], viewName, List.of());
```
WLefever-Cegeka, line 26: Would avoid this split and just create a getter on ViewSpecification for the attribute collectionName.

WLefever-Cegeka (commented):
Suggestion:
* update data model can be separated and PR'd to master
* maybe it's an idea to create an LdesConfig Bean from the new config that would allow to separately work on tasks as updating the fetching, ingestion, snapshotting

PR #541, opened 2023-04-24
Use mimetype for config input

Tomvbe:


ldes-server-port-admin-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/converters/LdesConfigModelConverter.java
```diff
@@ -46,8 +49,8 @@ public List<MediaType> getSupportedMediaTypes() {
 	@Override
 	protected LdesConfigModel readInternal(Class<? extends LdesConfigModel> clazz, HttpInputMessage inputMessage)
 			throws IOException, HttpMessageNotReadableException {
-		Model model = fromString(new String(inputMessage.getBody().readAllBytes(), StandardCharsets.UTF_8),
-				Lang.TURTLE);
+		Lang lang = getLang(Objects.requireNonNull(inputMessage.getHeaders().getContentType()), INGEST);
```
Yalz, line 52: This "INGEST" will provide extra context of the origin for when exception is thrown (when it's not a valid language).
Maybe provide a new ENUM value to distinguish between member ingestion and providing a new config?

Yalz (approved):

PR #542, opened 2023-04-24
Feature/extract shacle from ldesconfigmodel

jobulcke:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShape.java
```diff
@@ -0,0 +1,37 @@
...
+		return model;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o)
+			return true;
+		if (!(o instanceof ShaclShape that))
+			return false;
+		return Objects.equals(collection, that.collection);
```
Tomvbe, line 30: why are we only comparing the collection and ignoring the model? I noticed that on Wouter his branch, the model is included as well.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShape.java
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(collection);
```
Tomvbe, line 35: why are we only comparing the collection and ignoring the model? I noticed that on Wouter his branch, the model is included as well.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShape.java
```diff
@@ -0,0 +1,37 @@
...
+	public String getCollection() {
+		return collection;
+	}
+
+	public Model getModel() {
+		return model;
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
Tomvbe, line 25: can you add unit tests for the equals and hashcode? You can use a parameterized test for this. Cfr. ViewNameTest for a complete parameterized example or TokenTest for a hybrid example.

ldes-server-domain/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShapeTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+import java.util.stream.Stream;
+
+import static org.apache.jena.rdf.model.ResourceFactory.*;
+import static org.apache.jena.rdf.model.ResourceFactory.createResource;
+import static org.junit.jupiter.api.Assertions.*;
+
+class ShaclShapeTest {
```
Tomvbe, line 18: Since your tests do not fail while the model is not included in the equals and hashcode, your tests do not cover all cases :smile:. 

Can you also add tests for the hashcode?

You can check ViewNameTest for a complete example.

Tomvbe:
Can you remove LdesConfigModel.createLdesConfigShape method? I see that it is not used anymore after this refactor?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/InMemoryShaclCollection.java
```diff
@@ -1,42 +1,48 @@
...
+		return shapes.stream()
+				.filter(shaclShape -> shaclShape.getCollection().equals(collectionName))
+				.findFirst();
+	}
+
+	@EventListener
+	public void handleShaclChangedEvent(ShaclChangedEvent event) {
+		shaclShapeRepository.saveShaclShape(event.getShacl());
+		shapes.remove(event.getShacl());
+		shapes.add(event.getShacl());
```
Tomvbe, line 41: I assume this remove + add works because you only compare the collection in your equality/hashcode of your shaclShape. But if you fix that, this won't work anymore. You can fix this by using a map where the key is the collectionName

Tomvbe: also just doing 'add' without remove would have the same result as 'remove' + 'add' since duplicates cannot exist in a set and the old one would be replaced

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/InMemoryShaclCollection.java
```diff
@@ -1,42 +1,48 @@
...
-	public LdesConfigModel retrieveShape(String collectionName) {
-		return new LdesConfigModel(collectionName, shapes.get(collectionName));
+	public Optional<ShaclShape> retrieveShape(String collectionName) {
+		return shapes.stream()
+				.filter(shaclShape -> shaclShape.getCollection().equals(collectionName))
+				.findFirst();
+	}
+
+	@EventListener
+	public void handleShaclChangedEvent(ShaclChangedEvent event) {
```
Tomvbe, line 38: this method has a duplicate implementation of 'saveShape'. Can't you  replace the below code with `saveShape(event.getShacl())` ?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/InMemoryShaclCollection.java
```diff
@@ -1,42 +1,48 @@
...
+		shaclShapeRepository.saveShaclShape(event.getShacl());
+		shapes.remove(event.getShacl());
+		shapes.add(event.getShacl());
 	}
 
 	@PostConstruct
```
Tomvbe, line 44: In this setup, we kind of retrieve everything once and cache it indefinitely until we do an update through the application. A refresh REST call may be desirable in the future to retrieve all the latest shapes from the database in case some scripts or queries were done on the database directly instead of using the the REST API. This happens when bulk changes are desired.

ldes-server-port-admin-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/converters/ModelConverter.java
```diff
@@ -0,0 +1,51 @@
...
+	@Override
+	public List<MediaType> getSupportedMediaTypes() {
+		return List.of(MediaType.ALL);
+	}
+
+	@Override
+	protected Model readInternal(Class<? extends Model> clazz, HttpInputMessage inputMessage)
+			throws IOException, HttpMessageNotReadableException {
+		return RdfModelConverter.fromString(new String(inputMessage.getBody().readAllBytes(), StandardCharsets.UTF_8),
+				Lang.TURTLE);
```
Tomvbe, line 38: Can you use mime type here? Check other http converters for example

Tomvbe:
can you also test the ShaclShapeMongoRepository

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShape.java
jobulcke: In spirit of DDD, ShaclShape is considered an entity and the collection is the entity's identity

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/shacl/entities/ShaclShape.java
jobulcke: In spirit of DDD, ShaclShape is considered an entity and the collection is the entity's identity

PR #550, opened 2023-04-26
Add application yaml to gitignore

Tomvbe:


Yalz (approved):
Best to just put it on all allocation yaml files instead of mentioning the folder structure

PR #573, opened 2023-05-04
Feature: Delete view from collection

WLefever-Cegeka:
Support to delete a view from a collection. This will update the viewrepository and disable fragmentation and retention processes of that view.
Next steps:
* go over all the TODO's, highest priority = side effects of adding/deleting a view such as creating a root node for that view, refragmenting all existing members of the collection, deleting all fragments of a deleted view


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/FragmentationExecutorImpl.java
```diff
@@ -72,7 +73,14 @@ private LdesFragment retrieveRootFragmentOfView(ViewName viewName, Observation p
 
 	@EventListener
 	public void handleViewAddedEvent(ViewAddedEvent event) {
+		// TODO create root fragment and refragment all earlier members of collection
```
Yalz, line 76: Really not a big fan of adding an additional 2 TODOs in our code. This is why i would prefer to go to the v2 branch already

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/valueobject/ViewDeletedEvent.java
```diff
@@ -0,0 +1,15 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.domain.view.valueobject;
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.viewcreation.valueobjects.ViewName;
+
+public class ViewDeletedEvent {
```
Yalz, line 5: This COULD be a record. But i presume you have a reason why you prefer to not have it like that?

Tomvbe (approved):

PR #576, opened 2023-05-04
Feature: VSDSPUB-645: Refragmentation of members when view is added

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.stream.Stream;
+
+@Service
+public class RefragmentationServiceImpl implements RefragmentationService {
+	private final MemberRepository memberRepository;
+	private final ObservationRegistry observationRegistry;
+
+	public RefragmentationServiceImpl(MemberRepository memberRepository, ObservationRegistry observationRegistry) {
```
Tomvbe, line 17: Maybe we can use the new architecture instead of calling the deprecated memberRepository? We could ask the ingester for his members?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationService.java
```diff
@@ -0,0 +1,7 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.domain.ldesfragment.services;
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.ldesfragment.entities.LdesFragment;
+
+public interface RefragmentationService {
```
Tomvbe, line 5: What is 'refragmentation'? Is this for creating fragments for a new view? Is this to refragment existing views to get rid of empty/semi-empty views? Multiple interpretations are possible with this name. Perhaps we should add some javadoc to say what this does and does not do?

Tomvbe: Or a different name that has no ambiguity is possible as well.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+		this.memberRepository = memberRepository;
+		this.observationRegistry = observationRegistry;
+	}
+
+	// TODO refragmentation of old members might take a while. During this period
+	// new arrived members will not be fragmented.
+	// In the server v2 (horizontal scaling) we should solve this. For example by
```
Tomvbe, line 24: Isn't this server v2?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+	private final MemberRepository memberRepository;
+	private final ObservationRegistry observationRegistry;
+
+	public RefragmentationServiceImpl(MemberRepository memberRepository, ObservationRegistry observationRegistry) {
+		this.memberRepository = memberRepository;
+		this.observationRegistry = observationRegistry;
+	}
+
+	// TODO refragmentation of old members might take a while. During this period
+	// new arrived members will not be fragmented.
```
Tomvbe, line 23: What happens in the current setup if a new member is ingested while we are still processing this?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
WLefever-Cegeka: Yes, but we won't be able to do horizontal scaling and the dynamically updating of config in one go. Hence, the remark that in this PR. We handle the dynamically updates of a view.

(see also the next remark)

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
WLefever-Cegeka: That new member won't be fragmented in this view. Since the Refragmentation happens before the ViewCollection of the fragmentation process is updated.
That's actually what the remark is about: There's a time window in which you are busy refragmenting old members, but in which you won't process nor queue newly ingested members.

When we revise the fragmentation process, we should take this into account, but in my opinion it would be too much of a change to also include this in the PR. Moreover, since the discussing on horizontal scaling are still ongoing.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/RefragmentationServiceImpl.java
WLefever-Cegeka: Was doubting here, but then I thought it would be easier to now test it with the old MemberRepository and replace this code in: https://vlaamseoverheid.atlassian.net/browse/VSDSPUB-655

Tomvbe (approved):

PR #580, opened 2023-05-05
Feature/handle deletion of view

WLefever-Cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/repository/LdesFragmentRepository.java
```diff
@@ -27,4 +27,6 @@ Optional<LdesFragment> retrieveNonDeletedChildFragment(String viewName,
 	void incrementNumberOfMembers(String fragmentId);
 
 	List<LdesFragment> retrieveFragmentsOfView(String defaultViewName);
+
+	void removeLdesFragmentsOfView(String asString);
```
Tomvbe, line 31: variable name -> asString?

WLefever-Cegeka: Updated

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/FragmentationStrategyCollectionImpl.java
```diff
@@ -44,4 +47,10 @@ public void handleViewAddedEvent(ViewAddedEvent event) {
 		fragmentationStrategyMap.put(event.getViewName(),
 				fragmentationStrategyForView);
 	}
+
+	@EventListener
+	public void handleViewDeletedEvent(ViewDeletedEvent event) {
+		ldesFragmentRemover.removeLdesFragmentsOfView(event.getViewName());
```
Tomvbe, line 53: We should be cautious with this architecture. Currently we have a chain where we call remover.remove which calls repository.remove which calls mongoRepository.remove.
I wonder if the fragmentRemover could be listening to these events as well? Then it wouldn't be a simple pass through class. This method could just update his internals (the strategyMap).

WLefever-Cegeka: To be checked what the best approach is:
* handle events separately
* handle events in one go
Initial reasoning why the ViewDeletedEvent is handled in one go is because ViewAddedEvent is handled the same way. In that case however, there is an order that needs to be respected. We don't have that constraint in the ViewDeletedEvent

ldes-server-infra-mongo/mongo-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/fragment/LdesFragmentMongoRepository.java
```diff
@@ -99,12 +99,16 @@ public void incrementNumberOfMembers(String fragmentId) {
 	}
 
 	@Override
-	public List<LdesFragment> retrieveFragmentsOfView(String viewName) {
+	public Stream<LdesFragment> retrieveFragmentsOfView(String viewName) {
```
Tomvbe, line 102: I feel like we lose the power of Stream because we first do a 'findAll' to a list and then stream it. If I remember correctly, MongoRepository supports 'streamAllBy' as well. Then it manages a cursor for you and you are actually streaming the data from the database.
Note: this comment is for all the Stream changes in this PR

Tomvbe (approved):

ldes-server-infra-mongo/mongo-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/fragment/LdesFragmentMongoRepository.java
WLefever-Cegeka: Top comment!

PR #582, opened 2023-05-05
feat/ testing with viewservice

pj-cegeka:


ldes-server-port-admin-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/exceptionhandling/AdminRestResponseEntityExceptionHandler.java
```diff
@@ -32,6 +32,7 @@ protected ResponseEntity<Object> handleLdesShacleValidationException(
 		return handleException(ex, HttpStatus.BAD_REQUEST, request);
 	}
 
+	// Todo: remove?
```
WLefever-Cegeka, line 35: Correct! :-) Good catch!
Included it in #584

PR #587, opened 2023-05-07
Add ingest v2 to documentation

Tomvbe:
Limited documentation added for server v2 ingest. We will have to revise the entire document when all modules have been implemented. Including adding the visual architecture https://miro.com/app/board/uXjVMTYw9jY=/ when it is finished.


README.md
```diff
@@ -100,7 +101,8 @@ To enrich the server, certain Maven profiles can be activated:
...
 |-----------------------------------------|--------------------------|---------------------------------------------------------------------------------|-----------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------|
-| **HTTP Endpoints (Fetch/Ingestion)**    | http-ingest              | Enables a HTTP endpoint for to insert LDES members.                             | [HTTP configuration](#example-http-ingest-fetch-configuration)              | Endpoint:<br><br>- URL: /{ldes.collection-name}<br>- Request type: POST<br>- Accept: "application/n-quads", "application/n-triples" |
+| **HTTP Ingest v2**                      | http-ingest-v2           | Enables a HTTP endpoint to insert LDES members.                                 | [HTTP configuration](#example-http-ingest-fetch-configuration)              | Endpoint:<br><br>- URL: /{ldes.collection-name}<br>- Request type: POST<br>- Accept: "application/n-quads", "application/n-triples" |
```
Yalz, line 104: Is there a need to still call it ingest v2 if its on the epic branch?

Yalz (approved):

README.md
Tomvbe: The profile http-ingest still exists as you need it to seed the old 'members' collection which is still required for all other functionalities such as fragmentation, snapshotting, retention, etc.
I removed the v2 from the first column but kept it in the second column to make sure the profile names match. We will have to update those at the end of implementing server v2 when removing the old ingest.

PR #611, opened 2023-05-12
feat: create default pagination when new eventstream is created

jobulcke:
Add a default view to a collection when the collection is created

### DIFF
--- a/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/ViewServiceImpl.java
+++ b/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/ViewServiceImpl.java
@@ -47,18 +47,12 @@ public class ViewServiceImpl implements ViewService {
 	@Override
 	public void addDefaultView(String collectionName) {
 		ViewName defaultViewName = new ViewName(collectionName, DEFAULT_VIEW_NAME);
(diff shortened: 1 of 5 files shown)


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/ViewServiceImpl.java
```diff
@@ -30,12 +36,29 @@ public ViewServiceImpl(ViewRepository viewRepository, ApplicationEventPublisher
...
+			throw new DuplicateViewException(defaultViewName);
+		}
+
+		FragmentationConfig fragmentation = new FragmentationConfig();
+		fragmentation.setName(DEFAULT_VIEW_FRAGMENTATION_STRATEGY);
+		fragmentation.setConfig(DEFAULT_VIEW_FRAGMENTATION_PROPERTIES);
+
+		ViewSpecification defaultView = new ViewSpecification(defaultViewName, List.of(), List.of(fragmentation));
+		viewRepository.saveView(defaultView);
+		eventPublisher.publishEvent(new ViewAddedEvent(defaultView));
```
WLefever-Cegeka, line 61: implementation is ok, but maybe it's an idea to create a ViewSpecification based on the collection name and the default parameters in the EventStreamServiceImpl and use addView.
That would also avoid the duplication of `view.isPresent`, `viewRepository.saveView` and `publishEvent.`

WLefever-Cegeka (approved):

PR #612, opened 2023-05-15
fix/ use hostname in stream and shape iri and update tests

pj-cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/eventstream/http/services/EventStreamResponseConverter.java
```diff
@@ -19,15 +20,16 @@ public class EventStreamResponseConverter {
...
-	public EventStreamResponseConverter(ViewSpecificationConverter viewSpecificationConverter) {
+	public EventStreamResponseConverter(AppConfig appConfig, ViewSpecificationConverter viewSpecificationConverter) {
 		this.viewSpecificationConverter = viewSpecificationConverter;
+		hostname = appConfig.getHostName();
 	}
 
 	public EventStreamResponse fromModel(Model model) {
-		final String collection = getIdentifier(model, createResource(EVENT_STREAM_TYPE)).orElseThrow()
-				.replace(LDES, "");
+		final String collection = getIdentifier(model, createResource(EVENT_STREAM_TYPE)).orElseThrow().getLocalName();
```
Tomvbe, line 32: This can result in an NPE. `getLocalName` returns null when the resource is not a URI. I think it would be better to insert `.map(Resource::getLocalName)` before the `orElseThrow`. Then maybe throw a clear exception instead of generic one?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/eventstream/http/services/EventStreamResponseConverter.java
```diff
@@ -19,15 +20,16 @@ public class EventStreamResponseConverter {
 
 	public static final String CUSTOM = "http://example.org/";
 	public static final Property MEMBER_TYPE = createProperty(CUSTOM, "memberType");
+	private final String hostname;
 	private final ViewSpecificationConverter viewSpecificationConverter;
 
-	public EventStreamResponseConverter(ViewSpecificationConverter viewSpecificationConverter) {
+	public EventStreamResponseConverter(AppConfig appConfig, ViewSpecificationConverter viewSpecificationConverter) {
```
Tomvbe, line 26: Since we only need the hostname, and we don't expect to need more from the appConfig in the future (this object may even be removed), it may be better to replace this by `String hostName`. Then this class can be instantiated easier as well, as you only need a String instead of an entire AppConfig instance.

ldes-server-domain/src/test/resources/eventstream/streams/ldes-with-named-views.ttl
```diff
@@ -5,32 +5,33 @@
 @prefix dcterms: <http://purl.org/dc/terms/> .
 @prefix custom: <http://example.org/> .
 @prefix mobiliteit: <https://data.vlaanderen.be/ns/mobiliteit#>
```
Tomvbe, line 7: isn't there a `.` missing here at the end?

Tomvbe (approved):

PR #629, opened 2023-05-24
Feat/delete inmemory collections

jobulcke:


WLefever-Cegeka:
We'll park this until end of sprint.

PR #635, opened 2023-05-25
fix: broken openapi endpoint

Yalz:


pom.xml
```diff
@@ -117,8 +118,12 @@
             <dependency>
                 <groupId>org.springdoc</groupId>
                 <artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
```
Tomvbe, line 120: Do we still need this?

Tomvbe (approved):

PR #638, opened 2023-05-26
Feature: VSDSPUB-712: Implement Version Based Retention Policy

WLefever-Cegeka:
Provided an implementation of VersionBasedRetentionPolicy that:

- implements the RetentionPolicy interface
- returns false if the tested Member does not have the timestamp or versionOf attribute
- Otherwise: returns false for the amountOfMembersToKeep most recent members and true for the other members (which makes them a candidate for removal).

Note: There's possibly a more efficient implementation possible which only retrieves the members with the same versionOf attribute once, but I prefer this implementation because:

- it acts on the member (similar as defined in the spec)
- it fits in the current retention policy implementation
- retention is so far never a resource (memory/cpu/iops) critical process, so there's no big need for optimizations.

### DIFF
--- a/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldes/retentionpolicy/versionbased/ReverseTimeStampComparator.java
+++ b/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldes/retentionpolicy/versionbased/ReverseTimeStampComparator.java
@@ -1,0 +1,12 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.domain.ldes.retentionpolicy.versionbased;
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.tree.member.entities.Member;
+
+import java.util.Comparator;
+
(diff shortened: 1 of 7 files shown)


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldes/retentionpolicy/versionbased/ReverseTimeStampComparator.java
```diff
@@ -0,0 +1,12 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.domain.ldes.retentionpolicy.versionbased;
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.tree.member.entities.Member;
+
+import java.util.Comparator;
+
+public class ReverseTimeStampComparator implements Comparator<Member> {
```
Yalz, line 7: I'm a bit in dubio whether this should just be a "TimeStampComparator" and that you add a reverse() on call, or this approach. But since it works, i'm ok with this

Yalz (approved):

PR #640, opened 2023-05-26
Fix: Creating Retention Policies via the Admin API is Spec Compliant (Part 1)

WLefever-Cegeka:
Part 1 of using spec compliant definitions for the retention policies.

### DIFF
--- a/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/RetentionModelExtractor.java
+++ b/ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/RetentionModelExtractor.java
@@ -12,7 +12,7 @@ import java.util.List;
 import static be.vlaanderen.informatievlaanderen.ldes.server.domain.constants.RdfConstants.RETENTION_TYPE;
 
 @Component
-public class RetentionStatementsExtractor {
+public class RetentionModelExtractor {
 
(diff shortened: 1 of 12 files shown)


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/RetentionModelExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Statement;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static be.vlaanderen.informatievlaanderen.ldes.server.domain.constants.RdfConstants.RETENTION_TYPE;
+
+@Component
```
Yalz, line 14: Is there a need to have this a component? 
I see opportunities for static methods?

Yalz (approved):

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/RetentionModelExtractor.java
WLefever-Cegeka: I'm still in dubio on how the class will evolve when we have an ontology for fragmentations. For that reason I used a Component for the moment being.

PR #641, opened 2023-05-26
Feat/vsdspub 659/server dcat implemantation

jobulcke:


ldes-server-port-admin-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/controllers/OpenApiServerDcatController.java
```diff
@@ -0,0 +1,91 @@
...
+									@ExampleObject(value = """
+											[{"@type":["http://www.w3.org/ns/dcat#Catalog"],
+											"http://purl.org/dc/terms/title":[{"@value":"My LDES'es","@language":"en"}],
+											"http://purl.org/dc/terms/description":[{"@value":"All LDES'es from publiser X","@language":"en"}]},{"@id":"http://www.w3.org/ns/dcat#Catalog"}]
+											""")
+							})
+					}) Model dcat);
+
+	@ApiResponse(responseCode = "200")
+	@ApiResponse(responseCode = "404")
```
Tomvbe, line 87: I would not return a 404 on a DELETE as it does not makes a lot of sense. Delete should also be idempotent: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

Just delete it if it is present and don't do anything if it isn't. The person calling this endpoint just wants this resource to not exist anymore.

ldes-server-port-admin-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/controllers/IsIsomorphic.java
```diff
@@ -19,6 +21,10 @@ public static IsIsomorphic with(Model model) {
 		return new IsIsomorphic(model);
 	}
 
+	public static Model isIsomorphicWith(Model model) {
```
Tomvbe, line 24: Can you use the existing solution for this? This class is already an argumentMatcher which makes it possible to use it as argThat(IsIsomorphic.with(model)). The method you are creating here could be confusing out of context since this also is a Resultmatcher. If you see IsIsomorphic.isIsomorphicWith(model), without reading the actual code, no one would expect that there is a side-effect of 'argThat' being called.

When looking at the interface of this class it is also not clear what the difference is between IsIsomorphic.isIsomorphicWith(model) and IsIsomorphic.with(model). However they are very different.

ldes-server-infra-mongo/mongo-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/serverdcat/entity/ServerDcatEntity.java
```diff
@@ -0,0 +1,24 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.infra.mongo.serverdcat.entity;
+
+import org.springframework.data.annotation.Id;
+import org.springframework.data.mongodb.core.mapping.Document;
+
+@Document("server-dcat")
+public class ServerDcatEntity {
```
Tomvbe, line 7: Isn't this the catalog? -> DcatCatalogEntity?

Tomvbe:
Can you follow the same naming convention as done with the views and eventstreams? The servers, repos etc. all have a prefix Dcat. Ideally, it should not be visible that different people wrote the code of the different flows.

ldes-server-infra-mongo/mongo-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/infra/mongo/serverdcat/ServerDcatMongoRepository.java
```diff
@@ -0,0 +1,39 @@
...
+
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.serverdcat.entities.ServerDcat;
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.serverdcat.repositories.ServerDcatRepository;
+import be.vlaanderen.informatievlaanderen.ldes.server.infra.mongo.serverdcat.repository.ServerDcatEntityRepository;
+import be.vlaanderen.informatievlaanderen.ldes.server.infra.mongo.serverdcat.service.ServerDcatEntityConverter;
+
+import java.util.List;
+import java.util.Optional;
+
+public class ServerDcatMongoRepository implements ServerDcatRepository {
```
Tomvbe, line 11: ServerDcatMongoRepository -> DcatServerMongoRepository

Tomvbe (changes requested):
cfr comments

Tomvbe (approved):

PR #662, opened 2023-05-31
Feat/dataset fetch identity

pj-cegeka:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/eventstream/services/EventStreamServiceImpl.java
```diff
@@ -34,9 +40,10 @@ public List<EventStreamResponse> retrieveAllEventStreams() {
 		return eventStreamCollection.retrieveAllEventStreams().stream().map(eventStream -> {
 			List<ViewSpecification> views = viewService.getViewsByCollectionName(eventStream.getCollection());
 			ShaclShape shaclShape = shaclShapeService.retrieveShaclShape(eventStream.getCollection());
+			Optional<DcatDataset> dataset = dcatDatasetService.retrieveDataset(eventStream.getCollection());
 			return new EventStreamResponse(eventStream.getCollection(), eventStream.getTimestampPath(),
 					eventStream.getVersionOfPath(), eventStream.getMemberType(), eventStream.isDefaultViewEnabled(),
-					views, shaclShape.getModel());
+					views, shaclShape.getModel(), dataset.orElse(null));
```
Yalz, line 46: If I understand this correctly, this will include a jena model in the Response that we provide at the controller? This, together with the shape, don't look too ideal. No need to fix it immediately, but this should be on our radar to avoid

ldes-server-domain/src/test/resources/eventstream/streams/ldes-and-dataset-with-named-views.ttl
```diff
@@ -0,0 +1,70 @@
...
+@prefix dcat:  <http://www.w3.org/ns/dcat#> .
+@prefix foaf:  <http://xmlns.com/foaf/0.1/> .
+@prefix org:   <http://www.w3.org/ns/org#> .
+@prefix legal: <http://www.w3.org/ns/legal#> .
+@prefix m8g:   <http://data.europa.eu/m8g/> .
+@prefix locn:  <http://www.w3.org/ns/locn#> .
+@prefix view1:           <http://localhost:8080/collectionName1/view1/> .
+@prefix view2:           <http://localhost:8080/collectionName1/view2/> .
+
+server:collectionName1 a ldes:EventStream, dcat:Dataset ;
```
Yalz, line 21: So i've received the feedback from @WLefever-Cegeka that our application couldn't handle configuring collections and views in one request. I suppose this proves the opposite?

Yalz (approved):

ldes-server-domain/src/test/resources/eventstream/streams/ldes-and-dataset-with-named-views.ttl
Tomvbe: This is possible, you can create them at the same time and fetching the config of an eventstream also returns the view configs

Tomvbe (approved):

PR #666, opened 2023-06-01
Feat/vsdspub 691/request dcat

Tomvbe:
The user can request the complete and validated DCAT model through the admin API

The user and metadata crawlers can retrieve the complete and validated DCAT model through the public API


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/dcatserver/entities/DcatServer.java
```diff
@@ -21,6 +39,48 @@ public Model getDcat() {
...
+			List<DcatDataset> datasets) {
+		final List<Statement> statements = new ArrayList<>();
+		statements.addAll(createCatalogStatements(hostName));
+		statements.addAll(createDcatServiceStatements(hostName, dcatViews));
+		statements.addAll(createDcatDataSetStatements(hostName, datasets));
+		return statements;
+	}
+
+	private List<Statement> createCatalogStatements(String hostName) {
+		Resource dataServiceId = getDcat().listSubjectsWithProperty(RDF.type, DCAT_CATALOG).next().asResource();
```
Yalz, line 52: Is there no risk of no next() existing (no message sonar)?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/dcatserver/repositories/DcatServerRepository.java
```diff
@@ -13,4 +13,6 @@ public interface DcatServerRepository {
 	DcatServer saveServerDcat(DcatServer dcatServer);
 
 	void deleteServerDcat(String id);
+
+	Optional<DcatServer> findSingleDcatServer();
```
Yalz, line 17: is there a need to say "single" ? I think we use find[entity]() for most of our calls to get one item?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/dcatserver/services/DcatServerServiceImpl.java
```diff
@@ -1,21 +1,73 @@
...
+	private final DcatDatasetService dcatDatasetService;
+	private final String hostName;
+
+	private final DcatShaclValidator dcatShaclValidator;
 
-	public DcatServerServiceImpl(DcatServerRepository dcatServerRepository) {
+	public DcatServerServiceImpl(DcatServerRepository dcatServerRepository,
+			DcatViewService dcatViewService,
+			DcatDatasetService dcatDatasetService,
+			@Value("${ldes-server.host-name}") String hostName,
```
Yalz, line 35: Out of scope, but I'd like to get this out of the app config. So that this can still be changed on the fly :)

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/exceptions/LdesShaclValidationException.java
```diff
@@ -1,16 +1,25 @@
 package be.vlaanderen.informatievlaanderen.ldes.server.domain.exceptions;
 
+import org.apache.jena.rdf.model.Model;
+
 public class LdesShaclValidationException extends RuntimeException {
 
 	private final String validationReport;
+	private final Model validationReportModel;
```
Yalz, line 8: when/where are you using the model?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/validation/DcatShaclValidator.java
```diff
@@ -28,7 +31,13 @@ public boolean supports(Class<?> clazz) {
...
-		Model model = (Model) target;
-		validateShape(model);
+		if (isValidatorActive()) {
+			Model model = (Model) target;
+			validateShape(model);
+		}
+	}
+
+	private boolean isValidatorActive() {
+		return StringUtils.isNotBlank(dcatShapeUri);
```
Yalz, line 41: Does not blank also cover not null?

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/DcatViewService.java
```diff
@@ -4,6 +4,7 @@
 import be.vlaanderen.informatievlaanderen.ldes.server.domain.viewcreation.valueobjects.ViewName;
 import org.apache.jena.rdf.model.Model;
 
+import java.util.List;
 import java.util.Optional;
 
 public interface DcatViewService {
```
Yalz, line 10: I know it can be a standard of having Services and Repositories.. but it looks a bit like useless to have a Service that does little more than just a pass through.

Yalz (approved):

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/validation/DcatShaclValidator.java
Tomvbe: Checks if a CharSequence is not empty (""), not null and not whitespace only.
Whitespace is defined by Character.isWhitespace(char).
  StringUtils.isNotBlank(null)      = false
  StringUtils.isNotBlank("")        = false
  StringUtils.isNotBlank(" ")       = false
  StringUtils.isNotBlank("bob")     = true
  StringUtils.isNotBlank("  bob  ") = true

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/view/service/DcatViewService.java
Tomvbe: It does some minor mapping, sometimes an existence check. 

Personally I don't mind this setup as it is consistent across different flows. You know where to look for what and have a place to add logic when required. It also improves maintainability as I believe the repository should have no logic at all (except for fetching the entities and mapping them to value objects).
Sometimes calling the repository and sometimes the service from different places and for different methods would be less streamlined.

You can add this for discussion on the dev meeting if you want.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/exceptions/LdesShaclValidationException.java
Tomvbe: We have two flows. 
- The "/" endpoint that either returns the dcat or a http 500 without information.
- The "/admin/.../dcat" that either returns the dcat or a http 500 with a validation report. The validation report that we return here when status 500, is RDF.

ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/dcatserver/repositories/DcatServerRepository.java
Tomvbe: I did this because we do not provide an identifier. Usually you expect a single result when providing an identifier or a list when providing nothing. However when you expect just one result, it is common (in hibernate https://docs.jboss.org/hibernate/orm/4.3/javadocs/org/hibernate/result/ResultSetOutput.html#getSingleResult()) to explicitly mention this.

However I feel like having an id for the dcatServer is actually very user unfriendly (there is no way to know the id after creating, so all the other changes become impossible without consulting the db manually for the id) and I plan to propose on the demo prep to change this. When we work without an id, this single will become useless as well.

PR #691, opened 2023-06-05
Fix/consolidate fragment removers

Yalz:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/ldesfragment/services/TreeNodeRemoverImpl.java
```diff
@@ -42,6 +43,17 @@ public void removeTreeNodes() {
 						viewWithRetentionPolicies.getKey(), viewWithRetentionPolicies.getValue()));
```
Tomvbe, line 43: Didn't you remark that the 'removeTreeNodes' method is confusing as well because it removes treeNode references on members instead of actually removing tree nodes?

Tomvbe (approved):

PR #749, opened 2023-06-15
Make mongock config in application.yml optional

Tomvbe:


Yalz (approved):
Shame we need to do this, but clean solution !

PR #816, opened 2023-07-10
feat: rework fetch module to use AllocationRepository

Yalz:


ldes-fragmentisers/ldes-fragmentisers-common/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/repository/AllocationRepository.java
```diff
@@ -10,4 +12,6 @@ public interface AllocationRepository {
 	void unallocateAllMembersFromView(ViewName viewName);
 
 	void unallocateMembersFromCollection(String collectionName);
+
+	List<String> findMembersForFragment(String fragmentId);
```
Tomvbe, line 16: What does this return? Based on the name, I would expect Member objects but you return Strings? I assume these are ids? Maybe add this to the name of the method?

ldes-server-port-fetch-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/treenode/TreeNodeFactoryImpl.java
```diff
@@ -29,7 +34,13 @@ public TreeNode getTreeNode(LdesFragmentIdentifier treeNodeId, String hostName,
 		Fragment fragment = fragmentRepository.retrieveFragment(treeNodeId)
 				.orElseThrow(
 						() -> new MissingFragmentException(extendedTreeNodeId));
-		List<Member> members = memberRepository.getMembersByReference(treeNodeId.asString()).toList();
+
+		List<Member> members = allocationRepository.findMembersForFragment(treeNodeId.asString())
+				.stream()
+				.map(memberRepository::findById)
```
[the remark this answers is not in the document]
Tomvbe: Calling the memberRepository directly kind of goes against the modulith architecture as well.. As the api of the ingest module are the interfaces in be.vlaanderen.informatievlaanderen.ldes.server.ingest. Java does not allow it but everything inside the packages should be 'hidden' from other modules (spring-modulith implements it this way). I feel like the repository should stay property of the ingest module so that this can easily change without having to worry about other modules being impacted.

ldes-server-port-fetch-rest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/eventstream/valueobjects/EventStream.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @param timestampPath The ldes:EventStream instance MAY have a property ldes:timestampPath
+ * @param versionOfPath The ldes:EventStream instance MAY have a property ldes:versionOfPath
+ * @param shape         The ldes:EventStream instance SHOULD have a property tree:shape
+ */
+public record EventStream(String collection, String timestampPath, String versionOfPath, String shape,
+						  List<TreeNode> views) {
+
+	@Override
+	public boolean equals(Object obj) {
```
Tomvbe, line 17: I don't think this is necessary for a record?

ldes-server-port-fetch-rest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/rest/treenode/services/TreeNodeConverterImplTest.java
```diff
@@ -107,7 +103,7 @@ void when_TreeNodeHasMembersAndARelations_ModelHasMultipleStatements() {
 
 		Model model = treeNodeConverter.toModel(treeNode);
 
-		Assertions.assertEquals(9, getNumberOfStatements(model));
+		Assertions.assertEquals(10, getNumberOfStatements(model));
```
Tomvbe, line 106: Why is this now 10 instead of 9?

WLefever-Cegeka:
I'm not very involved in the refactor effort, but `Allocation` does not really look like an entity. 
Since it does not evolve over time, it might be worthwhile to consider converting it to a value object. 
Main question is: Should you be able to distinguish two allocations based on their properties or on their identity i.e. if two allocations are have the same memberId, the same viewName and the same fragmentId are they different things or not?
Additionally, I am doubting if the key `AllocationKey(String memberId, ViewName viewName)` is sufficient. A member could be part of multiple fragments in the same view (for example in two tiles in a geospatial view), so you would have multiple records with the same AllocationKey, which isn't possible.

pom.xml
```diff
@@ -370,10 +370,10 @@
 
         <plugins>
             <!-- CODE FORMATTER -->
-            <plugin>
-                <groupId>net.revelc.code.formatter</groupId>
-                <artifactId>formatter-maven-plugin</artifactId>
-            </plugin>
+<!--            <plugin>-->
```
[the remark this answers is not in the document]
Yalz: Already fixed :)

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/entities/Member.java
```diff
@@ -35,6 +35,13 @@ public String getId() {
 		return id;
 	}
 
+	public String getMemberIdWithoutPrefix() {
+		if (id.startsWith("http")) {
```
Tomvbe, line 39: wouldn't it be better to check if it starts with the collectionName?

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/entities/Member.java
```diff
@@ -35,6 +35,13 @@ public String getId() {
 		return id;
 	}
 
+	public String getMemberIdWithoutPrefix() {
```
Tomvbe, line 38: Can you add unit tests for this method?

Tomvbe (approved):

PR #841, opened 2023-07-19
Feat: Provide complete integration test for the retention module

WLefever-Cegeka:
Created an integration test for the retention module that acts on the captured events and tests the following policies:
* time-based
* version-based
* point-in-time

On a technical note: It is not possible to capture the outgoing MemberUnallocatedEvents and MemberDeletedEvents, because in the test set-up these are fired in a separate thread (due to the scheduling) and they cannot be captured with the spring test mechanism `ApplicationEvents`, since the test mechanism `ApplicationEventsApplicationListener` is not registered to that thread. Hence, the removal of members from views is verified using the database.

### DIFF
--- a/ldes-server-retention/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/retention/repositories/MemberPropertiesRepository.java
+++ b/ldes-server-retention/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/retention/repositories/MemberPropertiesRepository.java
@@ -8,8 +8,6 @@ import java.util.stream.Stream;
 
 public interface MemberPropertiesRepository {
 
-	// TODO Update implementation see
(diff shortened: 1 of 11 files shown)


ldes-server-retention/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/retention/repositories/MemberPropertiesRepository.java
```diff
@@ -8,6 +8,8 @@
 
 public interface MemberPropertiesRepository {
 
+	// TODO Update implementation see
```
Tomvbe, line 11: Can you remove this todo? We already have an issue for this

Tomvbe (approved):

PR #843, opened 2023-07-20
fix: broken conversion to LdesFragmentIdentifier

Yalz:


ldes-server-application/pom.xml
```diff
@@ -130,13 +130,23 @@
             </dependencies>
         </profile>
         <profile>
-            <id>fragmentation-pagination</id>
```
Tomvbe, line 133: why are you removing pagination?

ldes-server-application/pom.xml
```diff
@@ -130,13 +130,23 @@
...
+            <id>fragmentation</id>
             <dependencies>
+                <dependency>
+                    <groupId>be.vlaanderen.informatievlaanderen.vsds</groupId>
+                    <artifactId>ldes-fragmentisers-common</artifactId>
```
Tomvbe, line 137: shouldn't this be added to all fragmentation profiles?

ldes-server-application/pom.xml
```diff
@@ -130,13 +130,23 @@
...
+                    <version>${project.version}</version>
+                </dependency>
                 <dependency>
                     <groupId>be.vlaanderen.informatievlaanderen.vsds</groupId>
                     <artifactId>ldes-fragmentisers-pagination</artifactId>
                     <version>${project.version}</version>
                 </dependency>
+                <dependency>
```
Tomvbe, line 145: Shouldn't this be added to al fragmentation profiles?

Tomvbe:
Can we add a test that fails without this fix and succeeds with this fix? Somewhere in the integration test layer of the mongo repository?

ldes-server-application/pom.xml
Yalz: this actually isn't needed. since there already is a dependency in each module to common

Tomvbe (approved):

Issue #776, opened 2023-06-27
Instance http://localhost:8080/kbo/by-name?substring=& doesn't have a rdf:type tree:Node


xdxxxdx:
That is very specific bug, only occurs to the `tree:Node` `http://localhost:8080/kbo/by-name?substring=&`  for substring fragmentations.
I think it is an error when processing “&”。
This issue doesn't occur to other fragmented pages: e,g.

- http://localhost:8080/kbo/by-name?substring=v
- http://localhost:8080/kbo/by-name?substring=c

**Prerequitests:**

- LDES Server version: [ldes/ldes-server:1.1.0-SNAPSHOT](https://github.com/Informatievlaanderen/VSDS-LDESServer4J/issues/new?assignees=&labels=bug&projects=&template=bug_report.md&title=#)
- LDES Server is up and running on the port 8080

**Scenario:**
1.  Post KBO event stream configuration to the server endpoint by 

```
    @prefix ldes: <https://w3id.org/ldes#> .  
    @prefix custom: <http://example.org/> .  
    @prefix dcterms: <http://purl.org/dc/terms/> .  
    @prefix tree: <https://w3id.org/tree#>.  
    @prefix sh: <http://www.w3.org/ns/shacl#> .  
    @prefix server: <http://localhost:8080/> .  
    @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .  
    @prefix kbo: <http://localhost:8080/kbo/> .  
    @prefix legal: <http://www.w3.org/ns/legal#> .  
    @base <http://localhost:8080> .  
      
    server:kbo a ldes:EventStream ;  
    ldes:timestampPath dcterms:issued;  
    ldes:versionOfPath dcterms:isVersionOf ;  
    custom:memberType legal:legalEntity;  
    tree:shape kbo:shape .  
      
    kbo:shape a sh:NodeShape ;  
    sh:nodeShape [  
    sh:closed true ;  
    sh:propertyShape []  
    ] ;  
    sh:deactivated true .
```
2. Post the [BEL 20 data](https://github.com/xdxxxdx/pyldes_kbo/tree/main/sample/bel20) to the LDES Server
3. Post substring fragmentation view by 

```
    @prefix ldes: <https://w3id.org/ldes#> .  
    @prefix custom: <http://example.org/> .  
    @prefix dcterms: <http://purl.org/dc/terms/> .  
    @prefix tree: <https://w3id.org/tree#>.  
[...]

PR #850, opened 2023-07-25
fix: remove url character encoding within fragment

pj-cegeka:


pj-cegeka:
Issue https://github.com/Informatievlaanderen/VSDS-LDESServer4J/issues/776 will stay fixed because this only required the fragment url the server receives to be decoded. The only thing is that simply copying the url from a relation in a fragment won't work. This has to be encoded outside the server.

Tomvbe (approved):

PR #916, opened 2023-09-14
Implement first integration test: ingest small ldes

WLefever-Cegeka:
Notes:

- Integration test to ingest en fetch 1016 members
- In theory we don't need to convert evenstream to a model to post it later as a string, but I think in this case it was valuable to check first that it's a valid Jena Model. If it takes up too much resources, we can eliminate that step


Tomvbe (approved):
Good job Wouter!

PR #931, opened 2023-09-28
test: e2e: retention

Yalz:


ldes-server-integration-test/src/test/resources/features/retention.feature
```diff
@@ -0,0 +1,38 @@
+Feature: LDES Server Retention
+
+  @time-based
+  Scenario: Server provides timebased retention
+    Given I create the eventstream "data/input/eventstreams/mobility-hindrances_timebased.ttl"
+    When I ingest 30 members to the collection "mobility-hindrances"
+    Then the collection "mobility-hindrances" contains 30 members
+    When I wait for the retention cycle to take effect
+    # Since all added members' timestamp values equal to their ingestion date, they should be removed after 15 seconds
+    Then the collection "mobility-hindrances" contains 0 members
```
Lefeverw, line 10: I think it would be better to use awaitability here.

ldes-server-integration-test/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/LdesServerSteps.java
```diff
@@ -125,4 +142,16 @@ public void theResponseFromRequestingTheUrlHasAccessControlHeadersAndAnEtag(Stri
 		assertEquals("*", response.getHeader("Access-Control-Allow-Origin"));
 		assertNotNull(response.getHeader("ETag"));
 	}
+
+	@Then("the collection {string} contains {long} members")
+	public void theCollectionContainsMembers(String collection, long expectedMemberCount) {
+		long memberCountForCollection = memberRepository.getMemberStreamOfCollection(collection).count();
+
+		assertEquals(expectedMemberCount, memberCountForCollection);
+	}
```
Lefeverw, line 151: Since, these tests act on the boundaries of the ldes server, it would be better to fetch the page and verify it only contains 10 members instead of checking the database which is something internal that an external customer would not have access to.

Yalz: Was my first thought too.. yet the problem is.. Since we plan on making page Ids UUIDs, getting such page ids will be tricky..

Yalz: Although I follow your reasoning, I would make an exception for this. 
Retrieving the members of a collection would only be possible by first calling the view page, then follow its relations until we get members, ... only to then get the count. 
I think following the LDES itself is something the Client does, so I would take this shortcut for the integration tests on this level

WLefever-Cegeka (approved):

PR #956, opened 2023-10-30
Feat/668/shacl validator interface

jobulcke:


ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/domain/validation/DcatShaclValidator.java
```diff
@@ -11,33 +12,28 @@
...
+
+    @Override
+    protected void initializeShapes() {
+        if (isValidatorActive()) {
+            shapes = Shapes.parse(Objects.requireNonNull(dcatShapeUri));
+        }
+    }
+
+    @Override
+    public void validate(@NotNull Object target, @NotNull Errors errors) {
```
Tomvbe, line 29: Do we still ever use this part of the interface? It looks like we only use it for unit testing?

ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/domain/dcat/dcatserver/services/DcatServerServiceImpl.java
```diff
@@ -51,7 +51,7 @@ public Model getComposedDcat() {
 				.findSingleDcatServer()
 				.ifPresent(dcatServer -> composedDcat.add(getStatementsForComposedDcat(dcatServer)));
 
-		dcatShaclValidator.validate(composedDcat, null);
+		dcatShaclValidator.validate(composedDcat);
```
Tomvbe, line 54: Now you are calling the impl from the abstract class directly and bypassing the if check from the specific implementation:

```
        if (isValidatorActive()) {
            Model model = (Model) target;
            validate(model);
        }
```

ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/domain/validation/ShaclShapeValidator.java
```diff
@@ -14,14 +15,9 @@ protected void initializeShapes() {
 		shapes = Shapes.parse(FILE_NAME);
 	}
 
-	@Override
-	public boolean supports(Class<?> clazz) {
-		return Model.class.isAssignableFrom(clazz);
-	}
-
-	@Override
-	public void validate(Object target, Errors errors) {
```
Tomvbe, line 23: can we extract this implementation to the abstract parent class? Same for the other similar impls

Tomvbe (changes requested):
cfr. comments

ldes-server-admin/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/controllers/AdminViewsRestControllerTest.java
```diff
@@ -48,23 +49,23 @@
...
 	@Autowired
 	private ViewSpecificationConverter converter;
 	@Autowired
 	private MockMvc mockMvc;
 
-	@BeforeEach
-	void setUp() {
-		when(validator.supports(any())).thenReturn(true);
-	}
+//	@BeforeEach
```
Tomvbe, line 65: Can you remove this if you don't need it?

Tomvbe (approved):

PR #965, opened 2023-11-02
Feat/930/variable retention polling

jobulcke:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/constants/ServerConfig.java
```diff
@@ -26,6 +32,11 @@ public void setCompactionDuration(String compactionDuration) {
 		this.compactionDuration = compactionDuration;
 	}
 
+	public void setRetentionInterval(Integer retentionInterval) {
+		this.retentionInterval = retentionInterval;
+	}
+
 	public static final String HOST_NAME_KEY = "${ldes-server.host-name}";
```
Tomvbe, line 39: Why are these fields at the bottom of the class? Can you move them to the top?

README.md
```diff
@@ -261,6 +261,18 @@ Currently, there are 3 possible retention policies each with a different type an
     ] .
   ```
 
+##### Retention polling interval
+By default, every 10 seconds, the server checks if there are members that can be deleted that are not conform to the retention policy anymore.
+If a higher retention accuracy is desired, or a lower one if resources are limited for example, then a respectively lower or higher retention polling interval can be set in milliseconds.
```
Tomvbe, line 266: Can we support cron instead of ms-interval? This gives a lot more flexibility. Every 20 seconds would then look like this: "*/20 * * * * *". With the advantage that you can schedule it to run only at nights, or at weekends, once a month, whatever you desire.

Tomvbe: Unix usually supports a cron of 5 params, excluding seconds, however the spring @Scheduled adds a 6th param supporting seconds. It is important to clarify that in the readme here.

Tomvbe (changes requested):
cfr. comments

Tomvbe (approved):

Issue #998, opened 2023-11-20
Optimise Ldes Member Ingestion


Yalz:
**Describe the problem**
As part of performance optimisation, we noticed that the member ingestion algorithm performs two db queries per ingestion:
* verifies if member exists
* save member

This should be replaced with a single upsert query which can return a boolean wether the member was saved.

rorlic:
Be careful not to overwrite an existing member.

sandervd:
We don't need this, we need to create version objects in the server :trollface:

PR #1004, opened 2023-11-22
fix : Optimise Ldes Member Ingestion

larsic:
#998

avoid making two calls to db

adapt test

### DIFF
--- a/ldes-server-port-ingest/pom.xml
+++ b/ldes-server-port-ingest/pom.xml
@@ -55,6 +55,16 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback-classic.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
+            <version>${logback-core.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/ldes-server-port-ingest/pom.xml
+++ b/ldes-server-port-ingest/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>be.vlaanderen.informatievlaanderen.vsds</groupId>
         <artifactId>ldes-server</artifactId>
-        <version>2.3.0-SNAPSHOT</version>
+        <version>2.5.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>ldes-server-port-ingest</artifactId>
--- a/ldes-server-port-ingest/pom.xml
+++ b/ldes-server-port-ingest/pom.xml
@@ -55,16 +55,6 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-            <version>${logback-classic.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-core</artifactId>
-            <version>${logback-core.version}</version>
-        </dependency>
     </dependencies>
 
     <build>
--- a/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
+++ b/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
@@ -45,7 +45,7 @@ public class MemberIngesterImpl implements MemberIngester {
             eventPublisher.publishEvent(memberIngestedEvent);
             log.debug("Member with id {} ingested.", memberId);
         } else {
-            log.warn("Duplicate member ingested. Member with id {} already exist", member.getId());
+            log.warn("Duplicate member ingested. Member with id {} already exists", member.getId());
         }
     }
 
--- a/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
+++ b/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
@@ -36,9 +36,9 @@ public class MemberIngesterImpl implements MemberIngester {
     }
 
     private void ingestNewMember(Member member, String memberId) {
-        Optional<Member> savedMemberOpt = insert(member);
-        if (savedMemberOpt.isPresent()) {
-            Member savedMember = savedMemberOpt.get();
+        Optional<Member> memberSaved = insert(member);
+        if (memberSaved.isPresent()) {
+            Member savedMember = memberSaved.get();
             Metrics.counter("ldes_server_ingested_members_count").increment();
             final var memberIngestedEvent = new MemberIngestedEvent(savedMember.getModel(), savedMember.getId(),
                     savedMember.getCollectionName(), savedMember.getSequenceNr());
--- a/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
+++ b/ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
@@ -45,7 +45,7 @@ public class MemberIngesterImpl implements MemberIngester {
             eventPublisher.publishEvent(memberIngestedEvent);
             log.debug("Member with id {} ingested.", memberId);
         } else {
-            log.warn("Duplicate member ingested. Member with id {} already exists", member.getId());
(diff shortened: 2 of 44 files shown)


ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -26,30 +28,29 @@ public MemberIngesterImpl(MemberIngestValidator validator, MemberRepository memb
...
+        Optional<Member> savedMemberOpt = insert(member);
+        if (savedMemberOpt.isPresent()) {
+            Member savedMember = savedMemberOpt.get();
+            Metrics.counter("ldes_server_ingested_members_count").increment();
+            final var memberIngestedEvent = new MemberIngestedEvent(savedMember.getModel(), savedMember.getId(),
+                    savedMember.getCollectionName(), savedMember.getSequenceNr());
+            eventPublisher.publishEvent(memberIngestedEvent);
+            log.debug("Member with id {} ingested.", memberId);
+        } else {
+            log.warn("Duplicate member ingested. Member with id {} already exist", member.getId());
```
Yalz, line 48: typo : exists* 😉

ldes-server-port-ingest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImplTest.java
```diff
@@ -29,6 +36,7 @@
 @ExtendWith(MockitoExtension.class)
 class MemberIngesterImplTest {
 
+	public static final String DUPLICATE_MEMBER_INGESTED_MEMBER_WITH_ID_ALREADY_EXIST = "Duplicate member ingested. Member with id {} already exist";
```
Yalz, line 39: typo here again: exists* 😉

ldes-server-port-ingest/pom.xml
```diff
@@ -55,6 +55,16 @@
...
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback-classic.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
+            <version>${logback-core.version}</version>
+        </dependency>
```
Yalz, line 67: Are these deps required? Previously, we already had logging and didn't need these deps?

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -26,30 +28,29 @@ public MemberIngesterImpl(MemberIngestValidator validator, MemberRepository memb
...
+    @Override
+    public void ingest(Member member) {
+        validator.validate(member);
+        final String memberId = member.getId().replaceAll("[\n\r\t]", "_");
+        ingestNewMember(member, memberId);
+    }
+
+    private void ingestNewMember(Member member, String memberId) {
+        Optional<Member> savedMemberOpt = insert(member);
+        if (savedMemberOpt.isPresent()) {
```
Yalz, line 40: Based on that this is the only place we use the return of the method, I think it'll be much better to use a boolean rather than an optional (we lose time mapping it from an db entity to a Member)

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -26,30 +28,29 @@ public MemberIngesterImpl(MemberIngestValidator validator, MemberRepository memb
...
-}
+    @Override
+    public void ingest(Member member) {
+        validator.validate(member);
+        final String memberId = member.getId().replaceAll("[\n\r\t]", "_");
+        ingestNewMember(member, memberId);
+    }
+
+    private void ingestNewMember(Member member, String memberId) {
+        Optional<Member> savedMemberOpt = insert(member);
```
Yalz, line 39: General dev team guideline: we usually don't include "opt" as part of the variable name 😉

ldes-server-port-ingest/pom.xml
larsic: testing if message is logged

jobulcke: Sorry to interrupt, but I was wondering why you want to test the logging? In my opinion, I think it is sufficient to check whether the eventPublisher has been called and so the dependencies can be removed again :)

jobulcke (changes requested):

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
larsic: done, i assumed the mapping had a function (i know assumed)

ldes-server-port-ingest/pom.xml
larsic: ok overkill , it has been removed

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
larsic: as discussed, mapping is needed, reverted to Optional again

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -26,30 +31,30 @@ public MemberIngesterImpl(MemberIngestValidator validator, MemberRepository memb
...
+            Member sMember = savedMember.get();
+            Metrics.counter(LDES_SERVER_INGESTED_MEMBERS_COUNT).increment();
+            final var memberIngestedEvent = new MemberIngestedEvent(sMember.getModel(), sMember.getId(),
+                    sMember.getCollectionName(), sMember.getSequenceNr());
+            eventPublisher.publishEvent(memberIngestedEvent);
+            log.debug(MEMBER_WITH_ID_INGESTED, memberId);
+        } else {
+            log.warn(DUPLICATE_MEMBER_INGESTED_MEMBER_WITH_ID_ALREADY_EXISTS, memberId);
+        }
+    }
```
jobulcke, line 53: Maybe I am nitpicking here a little, but I see the public method only has two lines of code, where only the memberId as extracted and passed to the private function, so I would rewrite it a little.
```java
    @Override
    public void ingest(Member member) {
        validator.validate(member);
        final String memberId = member.getId().replaceAll("[\n\r\t]", "_");
	insert(member).ifPresentOrElse(
			insertedMember-> handleSuccessfullMemberInsertion(member, memberId),
			() -> log.debug(DUPLICATE_MEMBER_INGESTED_MEMBER_WITH_ID_ALREADY_EXISTS, memberId)
	);
    }

    private void handleSuccessfullMemberInsertion(Member member, String memberId) {
    	Metrics.counter(LDES_SERVER_INGESTED_MEMBERS_COUNT).increment();
	final var memberIngestedEvent = new MemberIngestedEvent(member.getModel(), member.getId(),
			member.getCollectionName(), member.getSequenceNr());
	eventPublisher.publishEvent(memberIngestedEvent);
	log.debug(MEMBER_WITH_ID_INGESTED, memberId);
    }
```
This way you can also see in the public method that we try to insert a new member in the repo and we can see what happens if it succeeds or fails

ldes-server-port-ingest/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImplTest.java
```diff
@@ -59,20 +60,18 @@ void whenValidatorThrowsAnException_thenTheIngestIsAborted_andTheExceptionIsThro
 
 	@Test
 	@DisplayName("Adding Member when there is a member with the same id that already exists")
-	void when_TheMemberAlreadyExists_thenMemberIsReturned() throws IOException {
+	void when_TheMemberAlreadyExists_thenFalseIsReturned() throws IOException {
```
jobulcke, line 63: Could you change the name of this test please, as we booleans are not used here anymore

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -26,30 +31,30 @@ public MemberIngesterImpl(MemberIngestValidator validator, MemberRepository memb
...
+
+    private Optional<Member> insert(Member member) {
```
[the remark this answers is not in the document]
larsic: it was changed to insertIntoRepo(Member)

jobulcke (approved):

PR #1021, opened 2023-11-30
Feat/metrics gauge micrometer gaugebuilder

pj-cegeka:


ldes-server-integration-test/src/test/resources/features/fetching.feature
```diff
@@ -37,4 +37,21 @@ Feature: Server basic fetching functionality
...
     And The response from requesting the url "/mobility-hindrances/paged?pageNumber=1" contains 0 remaining items statements
-    And I delete the eventstream "mobility-hindrances"
\ No newline at end of file
+    And I delete the eventstream "mobility-hindrances"
+
+  @clearRegistry
+  Scenario: Counter is created and returns number of inserted members
+    Given I create the eventstream "data/input/eventstreams/mobility-hindrances_paginated_1500.ttl"
+    When I ingest 1 members to the collection "mobility-hindrances"
+    Then The response from requesting the url "/actuator/prometheus" contains the message "ldes_server_ingested_members_count_total 1.0"
```
Yalz, line 46: If possible, I would write this as 
`The prometheus value for key "ldes_server_ingested_members_count_total" is 1`
Otherwise this BDD test seems very technical instead of being an outside view of the test.

PR #1022, opened 2023-11-30
docs: [skip ci] : Update README.md

jeflaenen:
Removed the reference to Swagger documentation since the reference referenced to nowhere.


Yalz (approved):
Server readme needs an extra look

PR #1025, opened 2023-12-04
Feat/vsdspub 1022/clear logging

jobulcke:


ldes-server-admin/pom.xml
```diff
@@ -33,6 +33,32 @@
             <artifactId>swagger-annotations-jakarta</artifactId>
         </dependency>
 
+        <!-- LOGGING -->
```
Yalz, line 36: Just a bit of nitpicking, let's call this "Monitoring and Logging" perhaps :) since i see prometheus is also part of this

ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/controllers/DcatViewsRestController.java
```diff
@@ -13,6 +14,7 @@
 import static org.apache.jena.riot.WebContent.contentTypeNQuads;
 import static org.apache.jena.riot.WebContent.contentTypeTurtle;
 
+@Observed
```
Yalz, line 17: Do Http Calls not already get observed on their own?

ldes-server-application/pom.xml
```diff
@@ -1,4 +1,25 @@
...
+        <!ELEMENT dependencies (dependency)*>
+        <!ELEMENT dependency (groupId|artifactId|version)*>
+        <!ELEMENT build (finalName|plugins)*>
+        <!ELEMENT finalName (#PCDATA)>
+        <!ELEMENT plugins (plugin)*>
+        <!ELEMENT plugin (groupId|artifactId)*>
+        <!ELEMENT profiles (profile)*>
+        <!ELEMENT profile (id|dependencies)*>
+        <!ELEMENT id (#PCDATA)>
+        ]>
```
Yalz, line 22: ? how is this needed?

ldes-server-application/pom.xml
```diff
@@ -24,10 +45,26 @@
...
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-aop</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-starter-webmvc-ui</artifactId>
```
Yalz, line 62: I see two spring starters.. this often means that dependencies are duplicated and perhaps not needed.
Can you tell me why you need the AOP dependency ? (what it brings to the application and why we need the starter)

ldes-server-application/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/logging/SimpleLoggingHandler.java
```diff
@@ -0,0 +1,34 @@
...
+	public void onError(@NotNull ObservedAspect.ObservedAspectContext context) {
+		log.atError().log(() -> getErrorInfo(context));
+	}
+
+	private String getErrorInfo(@NotNull ObservedAspect.ObservedAspectContext context) {
+		var joinPoint = context.getProceedingJoinPoint();
+		String problem = Objects.requireNonNull(context.getError()).getMessage();
+		String source = joinPoint.getSignature().getDeclaringType().getSimpleName();
+		String when = joinPoint.getSignature().getName();
+		return "ERROR - problem='%s', source='%s', when='%s'".formatted(problem, source, when);
```
Yalz, line 32: I think it might be safer to place the problem/error as last.. since this will be the biggest block

ldes-server-port-fetch-rest/pom.xml
```diff
@@ -52,6 +52,12 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+
+        <!-- LOGGING -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-aop</artifactId>
+        </dependency>
```
Yalz, line 60: We should normally only need the starter in the application. In here, we probably need the underlying libraries, no ?

ldes-server-port-ingest/pom.xml
```diff
@@ -55,6 +55,12 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
+
+        <!-- LOGGING -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-aop</artifactId>
```
Yalz, line 62: idem

Yalz (commented):
Looking good! Still missing some documentation however :)

ldes-server-port-fetch-rest/pom.xml
jobulcke: Yeah sorry, this was just me struggling a little with the dependencies and module-info.java. I will delete the dependency

ldes-server-port-ingest/pom.xml
jobulcke: Same answer

ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/rest/controllers/DcatViewsRestController.java
jobulcke: And the Http calls are also filtered out in the SimpleLoggingHandler, as the context from those calls are wrapped in a simple Observation.Context and we use the advanced ObservedAspect.ObservedAspectContext

Yalz (approved):

PR #1126, opened 2024-01-25
feat: dcat validity included in health endpoint

jobulcke:


docs/_configuration/dcat.md
```diff
@@ -21,5 +21,89 @@ For more info on DCAT, visit the [DCAT publication](https://www.w3.org/TR/vocab-
...
+        dcat-validity:
+          show-components: always
+          show-details: always
+          include: dcat
+```
+
+This config will first enable the `/actuator/health` endpoints. \
+Secondly, a mapping is provided, so when an `UNKNOWN`
+or `INVALID` status is returned, a http status 500 is associated with it. \
+As last, a group is defined, this ensures that only the details of the DCAT and its validity are exposed and the rest of
```
Tomvbe, line 74: As -> At

Tomvbe (approved):

PR #1148, opened 2024-02-06
feat: add default bucket to timebased fragmentation

pj-cegeka:


ldes-fragmentisers/ldes-fragmentisers-common/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/ModelParserTest.java
```diff
@@ -25,7 +25,7 @@ void setUp() throws URISyntaxException {
 	@Test
 	void when_MemberHasXMLDateTime_Then_ReturnLocalDateTime() {
 		LocalDateTime actual = ModelParser.getFragmentationObjectLocalDateTime(memberModel, ".*",
-				"http://purl.org/dc/terms/created");
+				"http://purl.org/dc/terms/created").get();
```
Tomvbe, line 28: immediately calling get from an optional is considered bad practice, can you change this variable to Optional<LocalDateTime> and then change the assertEquals below into: `assertThat(actual).contains(time);`. This assert4j logic checks for empty and the value in one operation.

Tomvbe (approved):

PR #1150, opened 2024-02-07
fix: handle malformed datetime

pj-cegeka:


ldes-fragmentisers/ldes-fragmentisers-common/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/ModelParser.java
```diff
@@ -49,7 +54,15 @@ public static List<LocalDateTime> getFragmentationObjectsLocalDateTime(Model mod
 				.filter(statement -> statement.getSubject().toString().matches(subjectFilter))
 				.map(Statement::getObject)
 				.map(RDFNode::asLiteral)
-				.map(literal -> localDateTimeConverter.getLocalDateTime((LiteralImpl) literal))
+				.map(literal -> {
+					try{
+						return localDateTimeConverter.getLocalDateTime((LiteralImpl) literal);
+					} catch (DatatypeFormatException exception) {
```
Tomvbe, line 60: This will work for DatatypeFormatException but what if we get an IllegalDateTimeFieldException in "parse.asCalendar()" ? Or what if the line above gives an exception, when our RDFNode is not a literal and someone put a resource in there instead? In the last case we will get a LiteralRequiredException from Jena.

The list can be endless, perhaps you should consider having a try catch somewhere earlier and more generic that allows you to fall back to the default bucket?

Tomvbe (changes requested):

Tomvbe (approved):

PR #1151, opened 2024-02-07
fix: [ci skip] document that cron is utc

Tomvbe:


docs/how-to-run.md
```diff
@@ -189,7 +189,7 @@ Here is an explanation provided for all the possibilities on how to tweak and co
 
 
 > **Note** ***: Unix usually supports a cron expression of 5 parameters, which excludes seconds. However, the spring
-> annotation `@Scheduled` adds a 6th parameter to support seconds.
+> annotation `@Scheduled` adds a 6th parameter to support seconds. Note that the cron schedules are in timezone 'UTC'.
```
jobulcke, line 192: I would rather write something like "Also be aware that the ...", because now you have a note in a note. But maybe this is just me nitpicking on something very small

jobulcke (approved):

Issue #1186, opened 2024-03-06
[Spec]: `tree:InBetweenRelation` should not be used as current way in time based fragmentations.


xdxxxdx:
[Spec]: `tree:InBetweenRelation` should not be used as the current way in time-based fragmentations.
```
@prefix prov:  <http://www.w3.org/ns/prov#> .
@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix terms: <http://purl.org/dc/terms/> .
@prefix tree:  <https://w3id.org/tree#> .

<../../bearb/timebased?year=2015>
        rdf:type        tree:Node;
        terms:isPartOf  <../../bearb>;
        tree:relation   [ rdf:type    tree:InBetweenRelation;
                          tree:node   <../../bearb/timebased?year=2015&month=08>;
                          tree:path   prov:generatedAtTime;
                          tree:value  "2015-08"^^<http://www.w3.org/2001/XMLSchema#gYearMonth>
                        ] .
```
This has been discussed since the very beginning of this usage. 
![image](https://github.com/Informatievlaanderen/VSDS-LDESServer4J/assets/15314807/4c136687-0782-4c6d-8304-a49a788d39f3)
@pietercolpaert addressed concern on this. https://github.com/TREEcg/specification/issues/82#issuecomment-1709605508 
https://vsdscore.slack.com/archives/C03CQKA4RU7/p1691138145204079

Thanks

Yalz:
AP: To replace with GreaterThanOrEquals and LesserThan relations pointing to each fragment

PR #1200, opened 2024-03-18
fix: #1186: update timebased relations

Yalz:


ldes-fragmentisers/ldes-fragmentisers-timebased-hierarchical/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentisers/timebasedhierarchical/services/TimeBasedRelationsAttributer.java
```diff
@@ -29,11 +29,16 @@ public TimeBasedRelationsAttributer(FragmentRepository fragmentRepository,
 
 	public void addInBetweenRelation(Fragment parentFragment, Fragment childFragment) {
 		FragmentationTimestamp timestamp = timestampFromFragmentPairs(childFragment);
-		TreeRelation parentChildRelation = new TreeRelation(config.getFragmentationPath(),
+		TreeRelation parentGteRelation = new TreeRelation(config.getFragmentationPath(),
 				childFragment.getFragmentId(),
-				timestamp.asString(), timestamp.getType(),
```
Tomvbe, line 34: I don't think we still need the method `getType` on  FragmentationTimestamp, we also won't need it on granularity anymore and we won't need the field `type` on Granularity anymore either. Can you remove all of these so that we don't have dangling/unused code?

Tomvbe (approved):

PR #1220, opened 2024-03-26
fix: unexpected empty shacl behavior

jobulcke:
fix: unexpected behavior when no shacl shape is provided or shacl shape is empty

### DIFF
--- a/ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/spi/EventStreamConverterImpl.java
+++ b/ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/spi/EventStreamConverterImpl.java
@@ -138,10 +138,14 @@ public class EventStreamConverterImpl implements EventStreamConverter {
     }
 
     private Model getShaclFromModel(Model model) {
-        final Statement shaclStatement = model.listStatements(null, TREE_SHAPE, (Resource) null).nextStatement();
-        List<Statement> shaclStatements = retrieveAllStatements(shaclStatement.getResource(), model);
-
-        return createDefaultModel().add(shaclStatements);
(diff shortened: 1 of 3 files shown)


ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/spi/EventStreamConverterImpl.java
```diff
@@ -138,10 +138,11 @@ private Optional<Boolean> getBooleanResource(Model model, Property predicate) {
...
     private Model getShaclFromModel(Model model) {
-        final Statement shaclStatement = model.listStatements(null, TREE_SHAPE, (Resource) null).nextStatement();
-        List<Statement> shaclStatements = retrieveAllStatements(shaclStatement.getResource(), model);
-
-        return createDefaultModel().add(shaclStatements);
+        final Model shaclModel = ModelFactory.createDefaultModel();
+        model.listStatements(null, TREE_SHAPE, (Resource) null)
+                .nextOptional()
+                .ifPresent(statement -> shaclModel.add(retrieveAllStatements(statement.getResource(), model)));
+        return shaclModel;
```
Yalz, line 145: this has the exact same behaviour as before.. what made you change this?

Yalz (approved):

ldes-server-admin/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/spi/EventStreamConverterImpl.java
jobulcke: If the <eventstream...> tree:shape [ TREE_SHAPE_INFORMATION ] is ommitted, then `.nextStatement` could be `null`, which could lead to a NullPointerException when trying to get the resource from that statement

PR #1234, opened 2024-03-29
fix: members deleted twice

jobulcke:
Before this fix, when an event stream was deleted, both the EventStreamDeletedEvent and ViewDeletedEvent were sent out, resulting that at some places in the code, both events were received and executed some logic twice


ldes-server-infra-mongo/mongo-admin-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/mongo/view/repository/DataServiceEntityRepository.java
```diff
@@ -2,6 +2,9 @@
 
 import be.vlaanderen.informatievlaanderen.ldes.server.admin.mongo.view.entity.DataServiceEntity;
 import org.springframework.data.mongodb.repository.MongoRepository;
+import org.springframework.data.mongodb.repository.Query;
 
 public interface DataServiceEntityRepository extends MongoRepository<DataServiceEntity, String> {
+    @Query(value = "{'_id':  {$regex:  ?0}}", delete = true)
```
pj-cegeka, line 8: This query can give problems if the name of a eventstream happens to contain the name of a different eventstream, for example: deleting the eventstream Observations would also delete all dcats of the views of eventstream QualityObservations.

pj-cegeka (changes requested):

pj-cegeka (approved):

PR #1237, opened 2024-04-02
Feat: improve blank node validation

pj-cegeka:


jobulcke (approved):
Looks good and I am happy with the result of the validator splitted in several parts 😄

Issue #1029, opened 2023-12-04
Actuator uses LDES server internal URLs


rorlic:
**Describe the bug**
When a LDES server is deployed behind a API gateway, the actuator uses the internal, private base URLs.

**To Reproduce**
Fetch https://telraam-api.net/ldes/actuator
Note that the internal addresses are being used.
Navigating to one of these links obviously does not work.

**Expected behavior**
The actuator uses the external addresses. The base address is configured in the `host-name`.

jobulcke:
Resolved in https://github.com/Informatievlaanderen/VSDS-LDESServer4J/pull/1246

PR #1246, opened 2024-04-05
fix: actuator uses internal addresses

jobulcke:
In https://github.com/Informatievlaanderen/VSDS-LDESServer4J/issues/1029, it was described that the actuator uses the internal addresses, which makes sense, as Spring Boot uses this internal address everywhere. This has now been fixed, where the property `ldes-server.host-name` is used to bind the internal address to this, which replaces the functionality of the already existing property `server.address`. 
This change comes with the precaution that the application will not startup when the server cannot be bound to `ldes-server.host-name`.

### DIFF
--- a/ldes-server-application/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/config/ActuatorHostNameCustomizer.java
+++ b/ldes-server-application/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/config/ActuatorHostNameCustomizer.java
@@ -1,0 +1,31 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.config;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.web.server.ConfigurableWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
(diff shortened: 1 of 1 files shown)


ldes-server-application/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/config/ActuatorHostNameCustomizer.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.web.server.ConfigurableWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
+import org.springframework.context.annotation.Configuration;
+
+import java.net.*;
+
+@Configuration
+public class ActuatorHostNameCustomizer implements WebServerFactoryCustomizer<ConfigurableWebServerFactory> {
+    private static final String HOST_NAME_KEY = "${ldes-server.host-name}";
```
Tomvbe, line 12: Can you use the constant from ServerConfig instead of duplicating this?

Tomvbe (approved):

PR #1268, opened 2024-04-17
docs: reduce documentation padding

ferre-vaes:
When the screen width exceeds 66.5rem, the padding is reduced to maximize the available screen real estate for the documentation content.

### DIFF
--- a/docs/_sass/_layout.scss
+++ b/docs/_sass/_layout.scss
@@ -125,9 +125,7 @@
     width: calc((100% - 90rem) / 2 + 16.5rem);
     min-width: 16.5rem;
   }
-}
 
-@media (min-width: 66.5rem) {
   .main {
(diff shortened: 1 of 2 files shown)


docs/_sass/_layout.scss
```diff
@@ -118,4 +118,17 @@
...
+}
+
+@media (min-width: 66.5rem) {
+  .side-bar {
+    width: calc((100% - 90rem) / 2 + 16.5rem);
+    min-width: 16.5rem;
+  }
+}
+
+@media (min-width: 66.5rem) {
```
jobulcke, line 130: I'm not such a (s)css expert, but can't those two block be merged into one?

jobulcke (approved):

docs/_sass/_layout.scss
ferre-vaes: Indeed, it is corrected

PR #1280, opened 2024-05-06
feat: postgres implementation

Yalz:


ldes-server-infra-postgres/postgres-fragmentation-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/converter/LdesFragmentIdentifierConverter.java
jobulcke, line 1: Where is this used? Because the conversion of the LdesFragmentIdentifier seems to happen already in the `FragmentEntity.fromLdesFragment` and `FragmentEntity.toLdesFragment`

ldes-server-infra-postgres/postgres-fragmentation-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/entity/FragmentEntity.java
```diff
@@ -0,0 +1,107 @@
...
+import be.vlaanderen.informatievlaanderen.ldes.server.domain.model.ViewName;
+import be.vlaanderen.informatievlaanderen.ldes.server.fragmentation.converter.FragmentPairConverter;
+import be.vlaanderen.informatievlaanderen.ldes.server.fragmentation.entities.Fragment;
+import jakarta.persistence.*;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+
+@SuppressWarnings({"java:S1068", "java:S107"})
```
jobulcke, line 15: A little suggestion, I don't know if we really want to use this, but I noticed that `java:S1068` is to suppress unused fields, but those fields can be annotated with `@Column` (without the need of specifying an additional name or etc.), and so could that specific warning suppression be dropped

ldes-server-infra-postgres/postgres-fragmentation-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/entity/FragmentEntity.java
```diff
@@ -0,0 +1,107 @@
...
+	private Boolean immutable;
+	private String parentId;
+	private Integer nrOfMembersAdded;
+	@ElementCollection(fetch = FetchType.EAGER)
+	@CollectionTable(
+			name = "fragmentation_fragment_relations",
+			joinColumns = @JoinColumn(name = "fragment_id")
+	)
+	private List<TreeRelationEntity> relations;
+	private String collectionName;
```
jobulcke, line 41: I know the field `collectionName` is copy pasted from the Mongo impl, but do we need to save the collectionName separately, as the viewName is already stored as `collectionName/viewName`, as `fragment.getFragmentId().getViewName().asString()` returns it in that format. 
I also don't see any usage or getter for the collectionName

ldes-server-infra-postgres/postgres-fragmentation-repository/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/sequence/FragmentSequenceRepositorySteps.java
jobulcke, line 1: Why not using a static import for the assertions?

ldes-server-infra-postgres/postgres-ingest-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/entity/MemberEntity.java
```diff
@@ -0,0 +1,71 @@
+package be.vlaanderen.informatievlaanderen.ldes.server.ingest.entity;
+
+
+import be.vlaanderen.informatievlaanderen.ldes.server.ingest.service.MemberEntityListener;
+import jakarta.persistence.*;
+
+import java.time.LocalDateTime;
+
+@Entity
+@EntityListeners(MemberEntityListener.class)
```
jobulcke, line 10: Do we still need this as the sequenceNr has now `@GeneratedValue`?

ldes-server-infra-postgres/postgres-retention-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/retention/projection/MemberPropertyVersionProjection.java
```diff
@@ -0,0 +1,24 @@
...
+	Set<MemberViewsProjection> getViews();
+	Integer getVersionNumber();
+
+	default MemberPropertiesEntity toMemberPropertiesEntity() {
+		return new MemberPropertiesEntity(getId(), getCollectionName(),
+				getViews().stream()
+						.map(MemberViewsProjection::getView)
+						.collect(Collectors.toSet()),
+				getVersionOf(), getTimestamp());
+	}
```
jobulcke, line 23: Could this be replaced to a mapper class or something, as default methods in interfaces are a bad practice and should only be used when dealing with legacy code

ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/MemberIngesterImpl.java
```diff
@@ -49,6 +49,7 @@ public boolean ingest(String collectionName, Model ingestedModel) {
 
         if (ingestedMembersCount != members.size()) {
             log.warn(DUPLICATE_MEMBERS_DETECTED);
+            log.info("HIT duplicate");
```
jobulcke, line 52: Is this a logging statement from debugging that is forgotten to be removed?

jobulcke (commented):
Please fix code smells

ldes-server-infra-postgres/postgres-fragmentation-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/entity/FragmentEntity.java
Yalz: The warning mentions unused private fields. which is still the case

ldes-server-infra-postgres/postgres-fragmentation-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/fragmentation/entity/FragmentEntity.java
Yalz: since we use this field for queries, we should keep it

ldes-server-infra-postgres/postgres-admin-repository/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/admin/postgres/dcatserver/DcatCatalogPostgresRepository.java
```diff
@@ -0,0 +1,53 @@
...
+	public void deleteServerDcat(String id) {
+		repository.deleteById(id);
+	}
+
+	@Override
+	public Optional<DcatServer> findSingleDcatServer() {
+		List<DcatServer> servers = repository.findAll().stream().map(converter::toDcatServer).toList();
+		if (servers.size() > 1) {
+			throw new IllegalStateException("Multiple dcatServers found. There should be 0 or 1 record of this.");
+		}
```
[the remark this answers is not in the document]
Yalz: Would prefer to tackle this later. Since this requires an update on the interface (so i'd do this if there is one implementation left.

PR #1316, opened 2024-06-17
Feat: new data model ingest

pj-cegeka:


ldes-server-infra-postgres/postgres-liquibase/src/main/resources/db/changelog/master.xml
jobulcke, line 1: I thought 3_0_1 also needed to be refactored into 3_1_0

Yalz (approved):
Further improvents will be added after this PR as discussed

PR #1351, opened 2024-08-02
feat: skolemization

jobulcke:


ldes-server-port-ingest/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/ingest/extractor/MemberExtractorDecorator.java
```diff
@@ -0,0 +1,19 @@
...
+
+	protected MemberExtractorDecorator(MemberExtractor baseMemberExtractor) {
+		this.baseMemberExtractor = baseMemberExtractor;
+	}
+
+	@Override
+	public List<IngestedMember> extractMembers(Model ingestedModel) {
+		return baseMemberExtractor.extractMembers(ingestedModel);
+	}
+}
```
Yalz, line 19: Would strongly advise against this. An abstract class that is just used in one class (just to avoid calling baseMemberExtractor.extractMembers(ingestedModel)) looks very overkill

Yalz (approved):

PR #1366, opened 2024-09-11
docs: include skolemization docs

Yalz:


docs/_features/skolemization.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+In this Skolemized version, the blank node has been replaced with the Skolem URI http://example.org/.well-known/genid/123456. 
+This URI is unique to the resource previously represented by the blank node, and can now be used to reference this 
+resource in other datasets. 
+This is a simple example, but it illustrates the basic process of Skolemization in Linked Data.
+
+## Skolemization in LDES Server
+
+To enable Skolemization, define the `<https://w3id.org/ldes#skolemizationDomain>` property when [setting up your Event Stream](../configuration/event-stream#configuring-a-new-event-stream) with for example with value `"http://example.org"`.
```
jobulcke, line 54: It is very good this extensive explanation of skolemization is written here, however, when it comes to configuring the LDES, I would add this to the page provided in the URI, as it can get very confusing if all the config props are spread out. 
I would go with the approach of the eventSource and putting there a link to this page instead of the other way around. (Or in both ways could also work 😉)

jobulcke (approved):

PR #1383, opened 2024-10-04
fix: broken compaction rework

jobulcke:


ldes-server-integration-test/src/test/java/be/vlaanderen/informatievlaanderen/ldes/server/LdesServerSteps.java
```diff
@@ -362,4 +363,10 @@ public void theResponseFromRequestingTheUrlDoesContainAJsonFile(String message,
 				.andReturn().getResponse();
 		assertTrue(response.getContentAsString().contains("%s{collection=\"%s\"} %s".formatted(message, collection, value)));
 	}
+
+	@And("the background processes did not fail")
+	public void theBackgroundProcessesDidNotFail() {
+		final List<String> failedStepExecutions = jdbcTemplate.queryForList("SELECT step_name FROM batch_step_execution WHERE status = 'FAILED'", String.class);
+		assertThat(failedStepExecutions).isEmpty();
+	}
```
Yalz, line 371: but .. this shouldn't matter if you get the correct results..? 
Sure we expect clean steps with no errors, but it doesn't really sound like a step i expect in a feature file of compaction

Yalz (approved):

ldes-server-maintenance/ldes-server-compaction/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/compaction/application/valueobjects/CompactedPages.java
```diff
@@ -18,7 +18,7 @@ public CompactedPages() {
 	}
 
 	public void addCandidatesToCompactedPages() {
-		if (!compactedCandidatesToAdd.isEmpty()) {
+		if (compactedCandidatesToAdd.size() > 1) {
```
Yalz, line 21: It can be just one page..?

PR #1402, opened 2024-10-23
chore: bump jena to next major version

jobulcke:


Yalz (approved):
Can't say I'm a fan of Jena doing this change..

Also double check spaces/tabs..

PR #1412, opened 2024-11-12
feat: add version in header

jobulcke:


ldes-server-domain/src/main/java/be/vlaanderen/informatievlaanderen/ldes/server/domain/versioning/AppVersionConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class AppVersionConfig {
+
+	@Bean
+	public String appVersion(BuildProperties buildProperties) {
+		return buildProperties.getVersion();
+	}
+}
```
Yalz, line 14: To me, this is too bloated.. Why not just use the build properties where needed? this created extra beans with no usage.

Yalz (changes requested):
Red build :)

Issue #1428, opened 2024-11-26
Not all members are bucketized for a by-page view when ingesting a larger data set using multiple pipelines


rorlic:
When running 2 pipelines to ingest data (AWV use case) and using a large enough data set (100K CSV lines), the by-page view does not contain all the members in the end, while both the by-location and by-time view do. The view stats show this clearly and it can be confirmed by counting the actual number of page members of the by-page view.

Because the ingested count is larger than the bucketized count and the bucketized count matches the paginated count, the problem clearly lies in the bucketization phase.

The problem cannot be reproduced it using a single pipeline at the time, so it must be a timing or concurrency issue.

PR #1430, opened 2024-11-26
fix: missing members when ingesting multi-threaded

rorlic:
fix: lock collection row to prevent interleaved sequence numbers in member_id when ingesting members


rorlic:
fixes github #1428

